PR #1, opened 2019-08-02
feat(CMake): Automatic configuration and installation of project Config.cmake.in file

panquez:


cmake/OpenGeode.cmake
```diff
@@ -57,20 +57,6 @@ install(
         .
 )
 
-#------------------------------------------------------------------------------------------------
-# Export and install OpenGeode configuration
-file(READ "${UTILS_FILE}" OPENGEODE_UTILS)
```
BotellaA: Leave the read of the utils file here

PR #11, opened 2019-08-19
Build/third parties

panquez:
Changing filesystem third party and create a wrapper to minizip features


cmake/OpenGeode.cmake
```diff
@@ -28,24 +28,20 @@ include("${UTILS_FILE}")
...
-)
-
 copy_windows_binaries(Async++)
 
 # Install OpenGeode third-parties
 install(
     DIRECTORY
         ${ASYNCPLUSPLUS_INSTALL_PREFIX}/
         ${BITSERY_INSTALL_PREFIX}/
+        ${MINIZIP_INSTALL_PREFIX}/
```
BotellaA: Minizip is still static ? If so, no need to install it

include/geode/georepresentation/io/detail/geode_boundary_representation_input.h
```diff
@@ -45,13 +37,9 @@ namespace geode
...
             : BRepInput( brep, std::move( filename ) )
         {
-            mz_zip_reader_create( &reader_ );
         }
 
-        ~OpenGeodeBRepInput()
-        {
-            mz_zip_reader_delete( &reader_ );
-        }
+        ~OpenGeodeBRepInput() {}
```
BotellaA: Are these destructors useful ?

include/geode/georepresentation/io/zip_file.h
```diff
@@ -0,0 +1,64 @@
+/*
```
BotellaA: Is this file in the right lib ? Is it related to basic, mesh or georepresentation ?

cmake/OpenGeode.cmake
panquez: Yes it is. Nice catch

include/geode/georepresentation/io/zip_file.h
panquez: It is a good question. I put it in georepresentation since it is in the only lib minizip was used for the moment. Maybe its place is in basic. What do you think of this?

BotellaA: I think basic might be a better place.

panquez: OK I move it into basic lib

include/geode/georepresentation/io/detail/geode_boundary_representation_input.h
panquez: Removed

cmake/utils.cmake
```diff
@@ -86,7 +86,7 @@ function(add_geode_library folder_path)
     )
     # TODO: Use TREE keyword when we change to cmake 3.8
```
BotellaA: You can also remove this commet since it is not true.

cmake/utils.cmake
```diff
@@ -73,7 +73,7 @@ function(add_geode_library folder_path)
     set(all_sources
         "${sources}"
         "${public_headers}"
-        "${private_headers}"
+        "${advanced_headers}"
```
BotellaA: You have to modify all the dependencies after this modifition.

PR #13, opened 2019-08-22
 fix(BRepOutput): add feature allowing other classes to archive BRep core files

panquez:


BotellaA:
If we start exporting this class, maybe it should be done for all geode input output classes

PR #20, opened 2019-08-30
build(Warning): Add pedantic warning level

BotellaA:
More warnings is always better! ;)


BotellaA:
More warnings is always better! ;)

PR #32, opened 2019-10-01
fix(Singleton): delay initialization of Singleton

BotellaA:


include/geode/basic/singleton.h
```diff
@@ -23,9 +23,11 @@
 
 #pragma once
 
+#include <memory>
```
panquez: Useful in .h ? If not, move it in the .cpp

PR #45, opened 2019-10-14
Update documentation to compile in MacOS with and without Xcode.

benjaminchauvin:


BotellaA:
Thanks a lot @benjaminchauvin for your PR !

PR #46, opened 2019-10-15
Ci/matrix

BotellaA:
Add build on ubuntu using g++ or clang++, on macOS using AppleClang or Xcode

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -9,7 +9,6 @@ on:
 jobs:
   format:
     runs-on: ubuntu-18.04
-    if: github.event_name == 'push'
     
     steps:
       - uses: actions/checkout@v1
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -23,6 +23,7 @@ jobs:
       matrix:
         cc: [gcc, clang]
         cxx: [g++, clang++]
+        config: [Debug, Release]
         exclude:
           - cc: gcc
             cxx: clang++
@@ -35,7 +36,7 @@ jobs:
       run: |
         mkdir -p build
         cd build
-        cmake -DCMAKE_BUILD_TYPE=Debug ..
+        cmake -DCMAKE_BUILD_TYPE=${{ matrix.config }} ..
         cmake --build . -- -j2
       env:
         CC: ${{ matrix.cc }}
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -37,26 +49,45 @@ jobs:
...
+      uses: pullreminders/slack-action@master
+      with:
+        args: '{\"channel\":\"CLJ1PS7UN\",\"text\":\"Compilation or Testing failed for OpenGeode on Ubuntu\"}'        
       
   build-mac:
     runs-on: macOS-10.14
     needs: format
+    strategy:
+      matrix:
+        cmake_args: ['-DCMAKE_BUILD_TYPE=Debug', '-G "Xcode"']
```
benjaminchauvin: Not sure that you need "" around Xcode.

.github/workflows/CI.yml
```diff
@@ -37,26 +49,45 @@ jobs:
...
     steps:
     - uses: actions/checkout@v1
     - name: Compile & Test
       run: |
         mkdir -p build
         cd build
-        cmake -DCMAKE_BUILD_TYPE=Debug ..
-        cmake --build . -- -j2
+        cmake ${{ matrix.cmake_args }} ..
+        cmake --build . --config Debug
```
benjaminchauvin: You do not need Debug when you use clang without Xcode. I presume that it does not hurt but is it possible to remove it when it is clang without Xcode (to be cleaner)?

.github/workflows/CI.yml
```diff
@@ -37,26 +49,45 @@ jobs:
...
         cd opengeode
-        ctest --output-on-failure
-      
+        ctest -C Debug --output-on-failure
```
benjaminchauvin: Same comment about the "debug".

benjaminchauvin:
There is no check in Release? It happens that it passes in debug and not in release.

.github/workflows/CI.yml
BotellaA: Yes I do. I tried without them first and it did not work

.github/workflows/CI.yml
BotellaA: I know but I did not find à clean way to do it yet... 😕

BotellaA:
No checks in release because it takes time and harder to implement. Not sure if it is really necessary. @panquez what do you think ?

benjaminchauvin:
> No checks in release because it takes time and harder to implement. Not sure if it is really necessary. @panquez what do you think ?

I do think that it is really important for a (surely more) very important reason: users will use the release version, not the debug one. Ensuring that the release works should be the priority. The debug version is just a tool for the developers. You should have both. This is just a personal opinion and I know that doing both complexifies everything (but for a good goal).

PR #54, opened 2019-10-24
add custom growable extension

### LINKED ISSUE #48: Enable backward compatibility

BotellaA:
fix #48


panquez:
LGTM but not easy to read all this PR. Maybe I do not see some mistakes. The better thing to do is to see if kcov passes through all serialize functions to see if they are all tested

BotellaA:
I will remove serialize for corners, surfaces, blocks and lines. How should we test the others ? Just à dummy test that save and load these struct ?

PR #58, opened 2019-10-29
Fix/copy vertex identifier

BotellaA:


include/geode/model/representation/builder/detail/copy.h
```diff
@@ -0,0 +1,268 @@
...
+
+namespace geode
+{
+    namespace detail
+    {
+        using Mapping = std::unordered_map< uuid, uuid >;
+
+        template < typename ModelFrom, typename ModelTo, typename Builder >
+        Mapping copy_corner_components(
+            const ModelFrom& from, const ModelTo& to, Builder& builder )
```
panquez: I think it is more readable calling "builder_to" or something like that

PR #79, opened 2019-11-25
fix(Copyright): add clear mention of modified files from RINGMesh

BotellaA:
@geologuic Are you ok with these modifications?


geologuic:
Thanks, @BotellaA. Looks good to me.

PR #85, opened 2019-12-05
feat(BoundingBox): add methods to compute the bounding box of meshes and models

panquez:


src/geode/model/representation/core/brep.cpp
```diff
@@ -467,4 +470,38 @@ namespace geode
...
+    {
+        BoundingBox3D box;
+        for( const auto& surface : surfaces() )
+        {
+            box.add_box( surface.mesh().bounding_box() );
+        }
+        if( nb_surfaces() > 0 )
+        {
+            return box;
+        }
```
BotellaA: Proposition:
add a template function taking a Range as parameter and then do 
if( nb_surfaces() > 0 ) return function( surfaces() )
if( nb_blocks() > 0 ) return function( blocks() )
if( nb_lines() > 0 ) return function( lines) )
return function( corners() )

PR #87, opened 2019-12-13
perf(Facets): optimize find and add facet

BotellaA:


include/geode/mesh/core/detail/facet_storage.h
```diff
@@ -87,21 +110,21 @@ namespace geode
...
                 {
                     counter_->set_value( id, counter_->value( id ) + 1 );
                     return id;
                 }
                 const auto size = facet_indices_.size();
-                facet_indices_[vertices] = size;
+                id = size;
                 facet_attribute_manager_.resize( size + 1 );
                 vertices_->modify_value(
                     size, SetValue{ vertices.vertices() } );
```
panquez:
```suggestion
                    id, SetValue{ vertices.vertices() } );
```
Same thing in return

PR #104, opened 2020-01-10
feat(Element deletion): returns old to new index mapping when deleting verticex, edges, facets, polygons or polyhedra in meshes

panquez:


include/geode/mesh/core/detail/facet_storage.h
```diff
@@ -159,6 +160,7 @@ namespace geode
                     facet_indices_.erase( key );
                 }
                 facet_attribute_manager_.delete_elements( to_delete );
+                return old2new;
```
BotellaA: Old2new is declare const. Not sure if it will force the copy Instead of the copy elision

PR #107, opened 2020-01-16
VertexIdentifier + PolygonalSurfaceBuilder

panquez:


src/geode/model/mixin/core/vertex_identifier.cpp
```diff
@@ -162,6 +186,67 @@ namespace geode
...
+            component_vertices_->modify_value( unique_vertex_id,
+                [&it]( std::vector< MeshComponentVertex >& vertices ) {
+                    vertices.erase( it );
+                } );
+        }
+
+        void update_unique_vertices( const ComponentID& component_id,
+            const std::vector< index_t >& old2new )
+        {
+            const auto unique_vertices_map =
```
BotellaA: not used

src/geode/model/mixin/core/vertex_identifier.cpp
```diff
@@ -162,6 +186,67 @@ namespace geode
...
+                                std::vector< MeshComponentVertex >& vertices ) {
+                                vertices.erase( it );
+                            } );
+                    }
+                    else
+                    {
+                        component_vertices_->modify_value( uv,
+                            [&it, &new_id](
+                                std::vector< MeshComponentVertex >& vertices ) {
+                                const index_t pos =
```
BotellaA:
```suggestion
                                const auto pos =
```

BotellaA:
@panquez maybe we should add a test for update_unique_vertices

PR #116, opened 2020-01-28
Feat/python binding

BotellaA:


bindings/python/src/basic/basic.cpp
```diff
@@ -0,0 +1,56 @@
...
+            : string_caster< absl::string_view, true >
+        {
+        };
+    } // namespace detail
+} // namespace pybind11
+
+PYBIND11_MODULE( OpenGeode_py_basic, module )
+{
+    module.doc() = "OpenGeode Python binding for basic";
+    module.attr( "NO_ID" ) = geode::NO_ID;
```
panquez: NO_ID is a string in python?

bindings/python/src/geometry/geometry.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include "point.h"
+#include "vector.h"
+
+namespace pybind11
+{
+    namespace detail
+    {
+        template < typename Type >
+        struct type_caster< absl::FixedArray< Type > >
+            : list_caster< absl::FixedArray< Type >, Type >
```
panquez: Is this needed for each module? Already define in basic.cpp

bindings/python/src/geometry/nn_search.h
```diff
@@ -0,0 +1,60 @@
...
+        .def( pybind11::init< std::vector< Point< dimension > > >() )          \
+        .def( "nb_points", &NNSearch##dimension##D::nb_points )                \
+        .def( "point", &NNSearch##dimension##D::point )                        \
+        .def( "closest_neighbor", &NNSearch##dimension##D::closest_neighbor )  \
+        .def( "radius_neighbors", &NNSearch##dimension##D::radius_neighbors )  \
+        .def( "neighbors", &NNSearch##dimension##D::neighbors )                \
+        .def( "colocated_index_mapping",                                       \
+            &NNSearch##dimension##D::colocated_index_mapping );                \
+                                                                               \
+    const auto name2##dimension =                                              \
```
panquez: name2 is not very nice

bindings/python/src/geometry/vector.h
```diff
@@ -0,0 +1,47 @@
...
+#define PYTHON_VECTOR( dimension )                                             \
+    const auto name##dimension = "Vector" + std::to_string( dimension ) + "D"; \
+    pybind11::class_< Vector##dimension##D, Point##dimension##D >(             \
+        module, name##dimension.c_str() )                                      \
+        .def( pybind11::init<>() )                                             \
+        .def( pybind11::init< std::array< double, dimension > >() )            \
+        .def( "length", &Vector##dimension##D::length )                        \
+        .def( "length2", &Vector##dimension##D::length2 )                      \
+        .def( "normalize", &Vector##dimension##D::normalize )                  \
+        .def( pybind11::self* double() )                                       \
```
panquez: and / ?

bindings/python/src/mesh/core/graph.h
```diff
@@ -0,0 +1,52 @@
...
+            .def( "edge_attribute_manager", &Graph::edge_attribute_manager,
+                pybind11::return_value_policy::reference )
+            .def( "edges_around_vertex", &Graph::edges_around_vertex );
+
+        pybind11::class_< EdgeVertex >( module, "EdgeVertex" )
+            .def( pybind11::init<>() )
+            .def( pybind11::init< index_t, index_t >() )
+            .def( pybind11::self == pybind11::self )
+            .def( pybind11::self != pybind11::self )
+            .def_readwrite( "edge_id", &EdgeVertex::edge_id )
```
panquez: what does it mean "_readwrite" ?

bindings/python/src/model/model.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include "representation/core/section.h"
+
+#include "representation/io/brep.h"
+#include "representation/io/section.h"
+
+namespace pybind11
+{
+    namespace detail
+    {
+        // template < typename Type, size_t dimension >
```
panquez: to remove

bindings/python/tests/basic/test-py-attribute.py
```diff
@@ -0,0 +1,106 @@
...
+    if len(manager.attribute_names()) != nb:
+        raise ValueError( "[Test] Not the correct number of attributes in the manager" )
+
+def test_delete_attribute_elements( manager ):
+    to_delete = [False] * manager.nb_elements()
+    to_delete[3] = True
+    to_delete[5] = True
+    manager.delete_elements( to_delete )
+    if manager.nb_elements() != len(to_delete) - 2:
+        raise ValueError( "[Test] Two attribute elements should have being removed" )
```
panquez:
```suggestion
        raise ValueError( "[Test] Two attribute elements should have been removed" )
```

bindings/python/tests/basic/test-py-attribute.py
```diff
@@ -0,0 +1,106 @@
...
+def test_sparse_attribute_after_element_deletion( manager ):
+    sparse_attribute = manager.find_attribute_double( "double" )
+    if sparse_attribute.value( 0 ) != 12:
+        raise ValueError( "Element 0 of sparse attribute should be 12 " )
+    if sparse_attribute.value( 5 ) != 7:
+        raise ValueError( "Element 5 of sparse attribute should be 7 " )
+    if sparse_attribute.value( 7 ) != 12:
+        raise ValueError( "Element 7 of sparse attribute should be 12 " )
+
+manager = basic.AttributeManager()
```
panquez: if __name__ == "__main__" ? To add in all the tests

bindings/python/src/basic/basic.cpp
BotellaA: No, it means it can be used with basic.NO_ID

bindings/python/src/mesh/core/graph.h
BotellaA: It means access in read an write to public data member from python

bindings/python/tests/basic/test-py-attribute.py
BotellaA: When this "if" could be false ? It is a test not a file to import. But I can add it by safety.

panquez: It means you can import one test in another to avoid declaration duplication. Maybe very rare but may be usefull. What do you think ?

bindings/python/src/geometry/geometry.cpp
BotellaA: yes it seems needed at compile time.

bindings/python/src/geometry/vector.h
BotellaA: removed * because already defined in Point

PR #120, opened 2020-01-29
feat(Attribute): add generic accessor converting value to float on th…

BotellaA:
…e fly


include/geode/basic/attribute.h
```diff
@@ -78,8 +78,40 @@ namespace geode
         absl::FixedArray< double > lambdas_;
     };
 
+    /*!     * Helper struct to convert an Attribute value to generic float.
```
panquez:
```suggestion
    /*! 
     * Helper struct to convert an Attribute value to generic float.
```

PR #123, opened 2020-02-07
fix(PImpl): fix macro usage outside geode namespace

BotellaA:


francoisbonneau:
I tryied with my plugin. it works. thanks

BotellaA:
No, macro should be outside the namespace, so they can be used anywhere.

PR #134, opened 2020-03-11
feat(PassKey): add passkey idiom instead of friendship

BotellaA:


src/geode/basic/attribute_manager.cpp
```diff
@@ -227,29 +232,29 @@ namespace geode
     void AttributeManager::register_attribute(
         std::shared_ptr< AttributeBase > attribute, absl::string_view name )
     {
-        return impl_->register_attribute( attribute, name );
+        return impl_->register_attribute( attribute, name, {} );
```
panquez: {} is for no key?

include/geode/basic/attribute_manager.h
```diff
@@ -113,7 +113,7 @@ namespace geode
                     "if an instantiated attribute of the same name "
                     "with different storage already exists." );
 
-                typed_attribute.reset( Attribute< T >::create( args... ) );
+                typed_attribute.reset( new Attribute< T >{ args..., {} } );
```
panquez: new ?

include/geode/basic/attribute.h
```diff
@@ -130,29 +138,31 @@ namespace geode
...
-        template < typename Archive >
-        void serialize( Archive& archive )
-        {
-            archive.ext( *this, DefaultGrowable< Archive, AttributeBase >{},
-                []( Archive& /*unused*/, AttributeBase& /*unused*/ ) {} );
-        }
+        virtual absl::string_view type() = 0;
 
-        virtual void resize( index_t size ) = 0;
+        virtual void resize( index_t size, AttributeKey /*unused*/ ) = 0;
```
panquez: Why adding the Key for all methods?

src/geode/basic/attribute_manager.cpp
BotellaA: is for creating the key

include/geode/basic/attribute_manager.h
BotellaA: yes, there is issue with make_shared since constructors are not trivials

include/geode/basic/attribute.h
BotellaA: thats all the methods that only the AttributeManager should be able to call

panquez: And functions are now in public?

BotellaA:
> I think I agree with the principle. But it is quite unusual to have at lot of unused parameters in class methods..

We can remove then, but it will probably be warnings

include/geode/basic/attribute.h
BotellaA: yes, public defines the callable methods and private the one not callable. With passkey you can restrict who can can the callables but in a sense there are callable from the outside.

include/geode/basic/passkey.h
```diff
@@ -0,0 +1,65 @@
...
+     *     };
+     *
+     *     Now, B class is the only class that can call A::restrictive_method
+     *     class B
+     *     {
+     *     public:
+     *          void run( A& a )
+     *          {
+     *              // the {} will implicitly creates a KeyForB
+     *              a.restrictive_method({});
```
panquez:
```suggestion
     *              // the {} will implicitly create a KeyForB
     *              a.restrictive_method( {} );
```

include/geode/basic/passkey.h
```diff
@@ -0,0 +1,65 @@
...
+
+namespace geode
+{
+    /*!
+     * This can be used to restrict which class can call methods
+     * Example:
+     *     class A
+     *     {
+     *         // Add a key only B can create
+     *         PASSKEY( B, KeyForB );
```
panquez: Do we have to show an example with two passkey or we assume it is clear enough?

include/geode/mesh/builder/edged_curve_builder.h
```diff
@@ -62,13 +66,17 @@ namespace geode
          */
         index_t create_point( const Point< dimension >& point );
 
+        void copy( const EdgedCurve< dimension >& edged_curve, BuilderKey )
+        {
+            copy( edged_curve );
```
panquez: why moving implementation in header file? To avoid BuilderKey naming?

include/geode/basic/passkey.h
BotellaA: Most usage will be will one passkey, if not just look at other code examples in OpenGeode

include/geode/mesh/builder/edged_curve_builder.h
BotellaA: no reason in particular. It is so simple I added it in th header by simplicity. I can move the code If you want.

panquez: No it's ok

PR #135, opened 2020-03-13
fix(Package): update path for pybind11 install.

francoisbonneau:


BotellaA:
The goal is to homogenize folders in the release package (and maybe fix a bug for free)

PR #140, opened 2020-03-20
Feat/absl span

BotellaA:


include/geode/basic/attribute.h
```diff
@@ -430,7 +430,7 @@ namespace geode
         void delete_elements( const std::vector< bool >& to_delete,
             AttributeBase::AttributeKey ) override
         {
-            std::vector< bool > to_keep( to_delete );
+            std::vector< bool > to_keep( to_delete.begin(), to_delete.end() );
```
BotellaA: Why this change?

panquez: It is a mistake when I try span< const bool >

BotellaA:
LGTM

PR #143, opened 2020-03-27
Feat/geometric computations

BotellaA:


bindings/python/src/geometry/geometry.cpp
```diff
@@ -24,9 +24,15 @@
 #include <pybind11/pybind11.h>
```
panquez: Missing header

BotellaA: No. There is a header! ;)

panquez: Oh my bad

PR #149, opened 2020-04-06
feat(VertexIdentifier): add methods to know if unique vertex stands for vertices in specific mesh component type or id

panquez:


panquez:
@BotellaA do you validate temporary changes in Ci and CD ?

PR #151, opened 2020-04-08
feat(Filename): add API function to extract filename (with ou with extension) from a path

panquez:


include/geode/basic/filename.h
```diff
@@ -0,0 +1,32 @@
...
+ *
+ */
+
+#pragma once
+
+#include <geode/basic/common.h>
+
+namespace geode
+{
+    std::string opengeode_basic_api filename_from_path(
```
BotellaA, line 30: I do not like function which can do several things. One function = One contract. Could you provide two functions? One for each contract?

What do you think?

panquez: I have no opinion on that question. I can do two functions. The major point for me is the return type. I would prefer absl::string_view but I do not see how to do it using ghc::filesystem::path

BotellaA: you cannot with filesystem, return type should own the data.

panquez: Ok. So std::string is good enough ?

BotellaA: yes.. until we move to C++17 for good

PR #152, opened 2020-04-10
feat(Solid): add some helper methods on micro-topology on Solid meshes

panquez:


src/geode/mesh/core/polyhedral_solid.cpp
```diff
@@ -500,6 +500,20 @@ namespace geode
         return get_polyhedron_vertex( polyhedron_vertex );
     }
 
+    template < index_t dimension >
+    index_t PolyhedralSolid< dimension >::vertex_in_polyhedron(
+        index_t polyhedron_id, index_t vertex_id ) const
+    {
+        for( const auto& v : Range{ nb_polyhedron_vertices( polyhedron_id ) } )
```
BotellaA, line 507:
```suggestion
        for( const auto v : Range{ nb_polyhedron_vertices( polyhedron_id ) } )

```

src/geode/mesh/core/polyhedral_solid.cpp
```diff
@@ -500,6 +500,20 @@ namespace geode
         return get_polyhedron_vertex( polyhedron_vertex );
     }
 
+    template < index_t dimension >
+    index_t PolyhedralSolid< dimension >::vertex_in_polyhedron(
```
BotellaA, line 504: absl::optinal?

panquez: why not? what is your opinion?

PR #157, opened 2020-04-15
fix(PolygonalSurface): fix polygon_edge_adjacent()

panquez:


src/geode/mesh/core/polygonal_surface.cpp
```diff
@@ -429,12 +429,14 @@ namespace geode
...
+            const auto edge_id = this->polygon_edge( polygon_edge );
             for( const auto e : Range{ nb_polygon_edges( polygon_adj ) } )
             {
                 const auto polygon = polygon_adjacent( { polygon_adj, e } );
+                const auto adj_edge_id =
+                    this->polygon_edge( { polygon_adj, e } );
                 if( polygon == polygon_edge.polygon_id
-                    && polygon_edge_vertex( { polygon_adj, e }, 0 ) == v1
-                    && polygon_edge_vertex( { polygon_adj, e }, 1 ) == v0 )
+                    && adj_edge_id == edge_id )
```
BotellaA, line 439: Maybe compute adj_edge_id directly in the if, so if the first test failed, it is not computed.

panquez: Done

PR #158, opened 2020-04-16
feat(Rotation): add rotation API function for Vector3D

panquez:


src/geode/geometry/rotation.cpp
```diff
@@ -0,0 +1,126 @@
...
+        const auto matrix = rotation_matrix( quaternion );
+        const Vector< 4 > quat_vector(
+            { vector.value( 0 ), vector.value( 1 ), vector.value( 2 ), 1.0 } );
+        Vector< 4 > result;
+        for( const auto i : geode::Range{ 4 } )
+        {
+            double coord{ 0. };
+            for( const auto j : geode::Range{ 4 } )
+            {
+                coord += matrix[i][j] * quat_vector.value( j );
```
BotellaA, line 115: Proposition: matrix could be defined as array<Vector<4>> and this operation could the resulat of a dot with the quaternion

src/geode/geometry/rotation.cpp
```diff
@@ -0,0 +1,126 @@
...
+namespace geode
+{
+    Vector3D opengeode_geometry_api rotate(
+        const Vector3D& vector, const Vector3D& axis, double angle )
+    {
+        if( axis.length2() < geode::global_epsilon2 )
+        {
+            OPENGEODE_EXCEPTION( axis.length2() < geode::global_epsilon2,
+                "[rotate] Given axis is almost null" );
+        }
```
BotellaA, line 92:
```suggestion
            OPENGEODE_EXCEPTION( axis.length2() < geode::global_epsilon2,
                "[rotate] Given axis is almost null" );
```

src/geode/geometry/rotation.cpp
panquez: Good idea

src/geode/geometry/rotation.cpp
```diff
@@ -0,0 +1,126 @@
...
+        matrix[3][3] = 1.0;
+
+        return matrix;
+    }
+} // namespace
+
+namespace geode
+{
+    Vector3D opengeode_geometry_api rotate(
+        const Vector3D& vector, const Vector3D& axis, double angle )
```
BotellaA, line 86: Could it be a Point3D we rotate around a Vector3D? Is there a reason to enforce a Vector?

PR #175, opened 2020-05-26
Fix/embedding

### LINKED ISSUE #166: Review naming between embedded and embedding

BotellaA:
fix #166

### DIFF
--- a/bindings/python/src/model/representation/core/brep.h
+++ b/bindings/python/src/model/representation/core/brep.h
@@ -32,8 +32,7 @@ namespace geode
         pybind11::class_< BRep, Topology, Corners3D, Lines3D, Surfaces3D,
             Blocks3D, ModelBoundaries3D >( module, "BRep" )
             .def( pybind11::init<>() )
-            .def(
-                "boundary_corners",
+            .def( "boundary_corners",
                 []( const BRep& brep, const Line3D& line ) {
                     std::vector< const Corner3D* > components;
                     for( const auto& component : brep.boundaries( line ) )
@@ -43,8 +42,7 @@ namespace geode
                     return components;
                 },
                 pybind11::return_value_policy::reference )
-            .def(
-                "boundary_lines",
+            .def( "boundary_lines",
                 []( const BRep& brep, const Surface3D& surface ) {
                     std::vector< const Line3D* > components;
                     for( const auto& component : brep.boundaries( surface ) )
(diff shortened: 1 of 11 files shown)


bindings/python/src/model/representation/core/brep.h
```diff
@@ -101,8 +95,7 @@ namespace geode
             .def( "nb_internal_surface_corners",
                 ( index_t( BRep::* )( const Surface3D& ) const )
                     & BRep::nb_internal_corners )
-            .def(
-                "internal_surface_corners",
```
mparquer: what about internal corners inside lines?
Also, thanks to the dimensions, we guess that the internal object is the corner but might be more obvious with internal_corners_inside_surface?

bindings/python/src/model/representation/core/brep.h
```diff
@@ -172,82 +161,76 @@ namespace geode
                     return components;
                 },
                 pybind11::return_value_policy::reference )
-            .def( "nb_embedded_corner_surfaces",
+            .def( "nb_embedding_corner_surfaces",
```
mparquer, line 175: again, thanks to the respective dimensions, we guess who is embedding who but it is not obvious to which object the adjective is related to. 
one suggestion : embedded_corner_inside_surfaces -> would give the corner
another suggestion: embedding_surfaces_of_corner -> would give the range of surfaces

bindings/python/src/model/representation/core/section.h
```diff
@@ -102,38 +96,35 @@ namespace geode
                     return components;
                 },
                 pybind11::return_value_policy::reference )
-            .def( "nb_embedded_corner_surfaces",
```
mparquer, line 105: what about embedded corner inside a line? maybe not different from an internal one

bindings/python/src/model/representation/core/brep.h
BotellaA: An internal corner in a ligne is just defining a boundary of two lines. It's our convention. What do you think?

bindings/python/src/model/representation/core/brep.h
BotellaA: @panquez an opinion on this? I agree that it is not obvious at all.

bindings/python/src/model/representation/core/brep.h
mparquer: I am ok if we can have make collections of lines. I didn't check.

bindings/python/src/model/representation/core/brep.h
panquez: The best name would be nb_embedding_surfaces. But since we cannot overload these functions we should specified on which type of objects they are applied. IMO, Marion's suggestion `embedding_surfaces_of_corner` is the best we can do

BotellaA: done

PR #176, opened 2020-05-27
Fix/no no

BotellaA:


src/geode/mesh/builder/polygonal_surface_builder.cpp
```diff
@@ -388,12 +387,13 @@ namespace geode
                 PolygonEdge edge{ polygon, e };
                 const auto edge_id = polygonal_surface_.edge_from_vertices(
                     { vertices_id[e], vertices_id[e + 1] } );
-                polygon_edges[edge_id].emplace_back( std::move( edge ) );
+                polygon_edges[edge_id.value()].emplace_back(
```
panquez, line 390: Should we add some ASSERT when using xxx.value()?

BotellaA: If there is no value inside an optional, accessing its value throw.

src/geode/mesh/core/polygonal_surface.cpp
```diff
@@ -434,34 +438,39 @@ namespace geode
...
-        return {};
+        return absl::nullopt;
     }
 
     template < index_t dimension >
     bool PolygonalSurfaceBase< dimension >::is_edge_on_border(
         const PolygonEdge& polygon_edge ) const
     {
-        return polygon_adjacent( polygon_edge ) == NO_ID;
+        return !polygon_adjacent( polygon_edge ).has_value();
```
panquez, line 473: `return !polygon_adjacent( polygon_edge )` is not enough?

src/geode/mesh/core/polyhedral_solid.cpp
```diff
@@ -1012,22 +1018,22 @@ namespace geode
...
-        return {};
+        return absl::nullopt;
     }
 
     template < index_t dimension >
     bool PolyhedralSolid< dimension >::is_polyhedron_facet_on_border(
         const PolyhedronFacet& polyhedron_facet ) const
     {
-        return polyhedron_adjacent( polyhedron_facet ) == NO_ID;
+        return !polyhedron_adjacent( polyhedron_facet ).has_value();
```
panquez, line 1036: same question

PR #179, opened 2020-05-30
Feat/mesh builder factory

BotellaA:


include/geode/mesh/builder/edged_curve_builder.h
```diff
@@ -84,13 +84,7 @@ namespace geode
             index_t vertex_id, const Point< dimension >& point ) = 0;
 
     private:
-        EdgedCurve< dimension >& edged_curve_;
+        EdgedCurve< dimension >* edged_curve_;
```
panquez, line 87: I suppose using smart_pointers is not possible (since set_mesh?)

BotellaA: smart_pointers are possible. However, this is a non-owning pointer... We could use weak_ptr but I do not like to use them.

PR #184, opened 2020-06-10
Feat/solid mesh

BotellaA:


src/geode/mesh/core/solid_mesh.cpp
```diff
@@ -0,0 +1,1093 @@
...
+            Facets::overwrite( from );
+        }
+
+        void overwrite_edges( const Edges& from )
+        {
+            Edges::overwrite( from );
+        }
+
+    private:
+        void convert_attribute_to_abseil()
```
panquez, line 326: should allow downward compatiblity? really cool

src/geode/mesh/core/solid_mesh.cpp
```diff
@@ -0,0 +1,1093 @@
...
+        return false;
+    }
+
+} // namespace
+
+namespace geode
+{
+    template < index_t dimension >
+    class SolidMesh< dimension >::Impl
+        : public detail::FacetStorage< PolyhedronFacetVertices >,
```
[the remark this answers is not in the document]
BotellaA: I left everything like it was regarding FacetStorage. I am not sure the potential perf improvement is forth the added complexity.

src/geode/mesh/core/solid_mesh.cpp
BotellaA: that's not new. It is an old convertion

PR #188, opened 2020-06-18
Feat/solid view

BotellaA:


include/geode/mesh/builder/tetrahedral_solid_view_builder.h
```diff
@@ -0,0 +1,96 @@
...
+    class TetrahedralSolidViewBuilder
+        : public TetrahedralSolidBuilder< dimension >
+    {
+        OPENGEODE_TEMPLATE_ASSERT_3D( dimension );
+
+    public:
+        static std::unique_ptr< TetrahedralSolidViewBuilder< dimension > >
+            create( TetrahedralSolidView< dimension >& mesh );
+
+        void add_viewed_vertex( index_t vertex_id );
```
panquez, line 53: I have not thinking of it before but what do you think about method like `add_viewed_vertices(absl::Span<const index_t> vertex_ids)` and `add_viewed_tetrahedra(absl::Span<const index_t> tetra_ids)`

BotellaA: maybe. It will not change the way it works behind the scene (adding one at the time)

panquez: Yes, you are right. The only question is where should it be done : in the class or in the client codes. I propose to keep that in mind but not apply changes right now (let's see how it is used in practice)

PR #195, opened 2020-06-22
fix(MeshBuilder): fix delete elements (wrong element around vertex update)

BotellaA:


src/geode/mesh/builder/surface_mesh_builder.cpp
```diff
@@ -469,6 +476,8 @@ namespace geode
         for( const auto v : Range{ surface_mesh_->nb_vertices() } )
         {
             to_delete[v] = !surface_mesh_->polygon_around_vertex( v );
+            DEBUG( surface_mesh_->polygon_around_vertex( v ).has_value() );
```
panquez, line 479: to delete

PR #197, opened 2020-06-24
Convert meshes into simplicial meshes

panquez:


src/geode/mesh/helpers/convert_solid_mesh.cpp
```diff
@@ -0,0 +1,99 @@
...
+            {
+                if( const auto adjacent =
+                        solid.polyhedron_adjacent( { p, e } ) )
+                {
+                    builder->set_polyhedron_adjacent(
+                        { p, e }, adjacent.value() );
+                }
+            }
+        }
+        return std::move( tet_solid );
```
BotellaA, line 92:
```suggestion
        return tet_solid;
```
Could you try this? Should use NRVO instead of move.

include/geode/mesh/helpers/convert_solid_mesh.h
```diff
@@ -0,0 +1,42 @@
...
+namespace geode
+{
+    FORWARD_DECLARATION_DIMENSION_CLASS( TetrahedralSolid );
+    FORWARD_DECLARATION_DIMENSION_CLASS( SolidMesh );
+} // namespace geode
+
+namespace geode
+{
+    template < index_t dimension >
+    absl::optional< std::unique_ptr< TetrahedralSolid< dimension > > >
```
BotellaA, line 39: What is the template for? Maybe remove it and export here.

src/geode/mesh/helpers/convert_solid_mesh.cpp
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    template absl::optional< std::unique_ptr< TetrahedralSolid3D > >
+        opengeode_mesh_api convert_solid_mesh_into_tetrahedral_solid(
+            const SolidMesh3D& );
```
BotellaA, line 97:
```suggestion
```
if you remove the template

PR #207, opened 2020-07-17
Feat/regular grid

BotellaA:


include/geode/mesh/io/regular_grid_input.h
```diff
@@ -0,0 +1,64 @@
...
+
+namespace geode
+{
+    FORWARD_DECLARATION_DIMENSION_CLASS( RegularGrid );
+} // namespace geode
+
+namespace geode
+{
+    /*!
+     * API function for loading an RegularGrid.
```
panquez, line 39:
```suggestion
     * API function for loading a RegularGrid.
```

src/geode/mesh/core/regular_grid.cpp
```diff
@@ -0,0 +1,234 @@
...
+            std::array< double, dimension > cells_size )
+            : origin_( std::move( origin ) ),
+              cells_number_( std::move( cells_number ) ),
+              cells_size_( std::move( cells_size ) )
+        {
+            index_t nb_cells{ 1 };
+            for( const auto c : cells_number )
+            {
+                nb_cells *= c;
+            }
```
panquez, line 50: It is really a detail point. In this kind of process, I always thinking of `std::accumulate` function. Which version is better using for instance `nb_cells = std::accumulate(cells_number.begin(), cells_number.end(), 1, std::multiplies<index_t>());` or the one you propose? I would say: your version is more clear, and maybe it is exactly the same for the compiler.

src/geode/mesh/core/regular_grid.cpp
```diff
@@ -0,0 +1,234 @@
...
+            return cells_number_.at( direction );
+        }
+
+        index_t cell_index( const Index& index ) const
+        {
+            index_t cell_id{ 0 };
+            for( const auto d : Range{ dimension } )
+            {
+                OPENGEODE_ASSERT( index[d] < cells_number_[d],
+                    "[RegularGrid::point] Invalid index" );
```
panquez, line 70:
```suggestion
                    "[RegularGrid::cell_index] Invalid index" );

```

src/geode/mesh/core/regular_grid.cpp
```diff
@@ -0,0 +1,234 @@
...
+            for( const auto d : Range{ dimension } )
+            {
+                const auto value =
+                    std::floor( ( query.value( d ) - origin_.value( d ) )
+                                / cells_size_[d] );
+                if( value < 0 || value >= cells_number_[d] )
+                {
+                    return absl::nullopt;
+                }
+                index[d] = static_cast< index_t >( value );
```
panquez, line 110: Should the cast be done after or before comparison with integer values?

tests/mesh/test-regular-grid.cpp
```diff
@@ -0,0 +1,108 @@
...
+void test_cell_number( const geode::RegularGrid3D& grid )
+{
+    OPENGEODE_EXCEPTION(
+        grid.nb_cells() == 750, "[Test] Wrong total number of cells" );
+    OPENGEODE_EXCEPTION(
+        grid.nb_cells( 0 ) == 5, "[Test] Wrong total number of cells" );
+    OPENGEODE_EXCEPTION(
+        grid.nb_cells( 1 ) == 10, "[Test] Wrong total number of cells" );
+    OPENGEODE_EXCEPTION(
+        grid.nb_cells( 2 ) == 15, "[Test] Wrong total number of cells" );
```
panquez, line 45: messages should be different for testing purpose

src/geode/mesh/core/regular_grid.cpp
BotellaA: If it is done before, you have issue because value can be negative

panquez: True!

PR #237, opened 2020-10-24
Fix/bitsery

BotellaA:
We need data to the new format for at least one test (test-convert-model-meshes)


panquez:
LGTM. As this PR is deeply breaking changes, I think this should be merged to master ASAP.

BTW, maybe you can add a test that save/load an object, and then add new registers e.g. in basic, then reload/save the object. It allows us to see that it is really working

panquez (commented):
Comments in main page of this PR

panquez (approved):

PR #238, opened 2020-10-25
fix(ModelInput): return the Model in the load_XX functions

BotellaA:
BREAKING CHANGE: parameters changed


BotellaA:
fix #232

panquez (approved):
Maybe some include (to brep and section) to remove

BotellaA:
> Maybe some include (to brep and section) to remove

No because we return the object.

PR #245, opened 2020-11-04
update upgrade guide

BotellaA:


panquez (approved):
Ok but add the info on exception raised if edges() are not enabled while used

PR #247, opened 2020-11-05
Feat/copy basic objects

panquez:


BotellaA (approved):
Should we do the same for the others objects?

panquez:
> Should we do the same for the others objects?

Yes we do

panquez (commented):
LGTM

PR #253, opened 2020-11-23
feat(Geometry): add distance between point and sphere

panquez:


src/geode/geometry/distance.cpp
```diff
@@ -540,6 +540,23 @@ namespace geode
         return std::make_tuple( std::fabs( distance ), projected_p );
     }
 
+    template < index_t dimension >
+    std::tuple< double, Point< dimension > > point_sphere_distance(
+        const Point< dimension >& point, const Sphere< dimension >& sphere )
+    {
+        Vector< dimension > center_to_point{ sphere.origin(), point };
+        if( center_to_point.length() < global_epsilon )
```
BotellaA, line 548:
```suggestion
        if( center_to_point.length() < sphere.radius() )
```

BotellaA (changes requested):

src/geode/geometry/distance.cpp
```diff
@@ -540,6 +540,23 @@ namespace geode
...
+        {
+            Vector< dimension > dummy_direction;
+            dummy_direction.set_value( 0, 1 );
+            return std::make_tuple( sphere.radius(),
+                sphere.origin() + dummy_direction * sphere.radius() );
+        }
+        return std::make_tuple(
+            std::fabs( center_to_point.length() - sphere.radius() ),
```
BotellaA, line 556:
```suggestion
            center_to_point.length() - sphere.radius(),
```

BotellaA (approved):

PR #254, opened 2020-11-24
fix(Geometry): add distance between point and Circle and Disk

panquez:


src/geode/geometry/distance.cpp
```diff
@@ -586,6 +586,80 @@ namespace geode
...
+                    global_epsilon ) ) // dummy_direction is aligned along plane
+                                       // normal
+            {
+                other_direction.set_value( 1, 1.0 );
+            }
+            OPENGEODE_ASSERT( !circle.plane().normal().inexact_equal(
+                                  other_direction, global_epsilon ),
+                "[point_circle_distance] Problem while getting circle nearest "
+                "point" );
+            const Vector3D other_projected_on_plane =
```
BotellaA, line 614:
```suggestion
            const auto other_projected_on_plane =
```
Is it working?

BotellaA (approved):

src/geode/geometry/distance.cpp
panquez: No, auto stands for Point3D that cannot be normalized()

PR #264, opened 2020-12-09
fix(SolidBuilder): update computation adjacency using Facets

BotellaA:


src/geode/mesh/builder/solid_mesh_builder.cpp
```diff
@@ -463,78 +463,36 @@ namespace geode
...
-                    const auto vertex =
-                        solid_mesh_->polyhedron_facet_vertex( { facet, v } );
-                    polyhedron_facet_around_vertex[vertex].push_back( facet );
-                }
+                PolyhedronFacet facet{ polyhedron, f };
+                const auto vertices_id =
+                    solid_mesh_->polyhedron_facet_vertices( facet );
+                const auto facet_id = facets.facet_from_vertices( vertices_id );
+                polyhedron_facets[facet_id.value()].emplace_back(
+                    std::move( facet ) );
```
panquez, line 481: Is it not faster to use a `absl::FixedArray< PolyhedronFacet >` instead of `absl::FixedArray< absl::InlinedVector< PolyhedronFacet, 2 > >` and if there is already the facet you set the adjacency?

panquez: And lighter?

panquez (approved):

PR #266, opened 2020-12-10
Fix/regular grid

panquez:


src/geode/mesh/core/regular_grid.cpp
```diff
@@ -332,6 +334,38 @@ namespace geode
         return impl_->cell_attribute_manager();
     }
 
+    template < index_t dimension >
+    BoundingBox< dimension > RegularGrid< dimension >::bounding_box() const
+    {
+        BoundingBox< dimension > bbox;
+        bbox.add_point( this->origin() );
+        Point< dimension > extreme_point;
+        for( const auto d : Range{ dimension } )
```
BotellaA, line 343: or first compute the Index and then use the method point( index ). The result will be the same

src/geode/mesh/core/regular_grid.cpp
```diff
@@ -332,6 +334,38 @@ namespace geode
...
+    template < index_t dimension >
+    RegularGrid< dimension > RegularGrid< dimension >::clone() const
+    {
+        std::array< index_t, dimension > cells_number;
+        std::array< double, dimension > cells_size;
+        for( const auto d : Range{ dimension } )
+        {
+            cells_number[d] = this->nb_cells( d );
+            cells_size[d] = this->cell_size( d );
+        }
```
BotellaA, line 362: You could add getters from the Impl to simplify.

BotellaA (approved):

PR #267, opened 2020-12-11
feat(LocalId): add new type local_index_t

BotellaA:
First porting to new type.


src/geode/mesh/core/surface_mesh.cpp
```diff
@@ -319,7 +355,8 @@ namespace geode
         const auto vertex = polygon_vertex.vertex_id;
         const auto polygon = polygon_vertex.polygon_id;
         const auto nb_vertices = nb_polygon_vertices( polygon );
-        return { polygon, ( vertex + 1 ) % nb_vertices };
+        return { polygon,
+            static_cast< local_index_t >( ( vertex + 1 ) % nb_vertices ) };
```
BotellaA, line 359: Cast are required when using arithmetic computation because unsigned char are promoted to int, so we need to narrow down the type.

tests/mesh/test-regular-grid.cpp
```diff
@@ -162,7 +162,7 @@ void test_clone( geode::RegularGrid3D& grid )
         "[Test] Clone missing attribute" );
     const auto clone_attribute =
         clone.cell_attribute_manager().find_attribute< int >( attribute_name );
-    for( const auto c : geode::Range{ clone.nb_cells() } )
+    for( const auto c : geode::TRange< int >{ clone.nb_cells() } )
```
panquez, line 165: Why TRange< int >? `nb_cells()` returns index_t, is not it?

panquez (approved):
Only a small question

panquez (approved):

PR #278, opened 2020-12-16
perf(SurfaceBuilder): improve surface mesh adjacency computation

BotellaA:


BotellaA:
Nope, this implementation forgets about non manifold...

panquez (approved):

PR #297, opened 2021-02-11
Feat/identifier

BotellaA:


include/geode/model/representation/io/detail/geode_brep_input.h
```diff
@@ -48,6 +49,11 @@ namespace geode
         void load_brep_files( absl::string_view directory )
         {
             BRepBuilder builder{ brep() };
+            builder.load_identifier( directory );
+            if( brep().name() == Identifier::DEFAULT_NAME )
```
BotellaA, line 53: Should we do the same for Mesh inputs?

include/geode/basic/identifier.h
```diff
@@ -0,0 +1,75 @@
...
+        void set_name( absl::string_view name, IdentifierKey );
+
+        void load_identifier( absl::string_view directory, IdentifierKey );
+
+    protected:
+        Identifier();
+
+        void set_id( uuid id );
+
+        void set_name( absl::string_view name );
```
panquez, line 66: useful?

include/geode/basic/identifier.h
```diff
@@ -0,0 +1,75 @@
...
+
+namespace geode
+{
+    class opengeode_basic_api Identifier
+    {
+        PASSKEY( IdentifierBuilder, IdentifierKey );
+        friend class bitsery::Access;
+
+    public:
+        static constexpr auto DEFAULT_NAME = "unknown";
```
panquez, line 44: Is "unknown" a good choice? Proposal: "default_name", "no_name", "unnamed", ...

panquez (approved):

include/geode/basic/identifier.h
BotellaA: yes, it is used in bitsery compatibility. But I could also use a Builder inside Bitsery if you prefer.

PR #299, opened 2021-02-15
feat/copy-relationships

BotellaA:
BREAKING CHANGE: ModelCopyMapping moved directly in geode namespace


panquez (approved):
Do not forget upgrade guide. Especially for the deletion of `copy_component_relationships` (replaced by `copy_relationships`)

PR #300, opened 2021-02-15
feat/relation-range

BotellaA:


src/geode/model/mixin/core/relationships.cpp
```diff
@@ -285,6 +285,17 @@ namespace geode
         impl_->unregister_component( id );
     }
 
+    index_t Relationships::nb_relations( const uuid& id ) const
+    {
+        return detail::count_relationships( relations( id ) );
```
panquez, line 290: `count_relationships` is costly. Could you directly use graph->edges_around()? Is it faster?

panquez (approved):

PR #309, opened 2021-03-03
fix(Geometry): fix incoherence of outputs of point_triangle_distance

panquez:


src/geode/geometry/distance.cpp
```diff
@@ -204,7 +204,6 @@ namespace geode
         auto det = std::fabs( a00 * a11 - a01 * a01 );
```
BotellaA, line 204: I think `c` above is not used anymore

BotellaA (approved):

PR #322, opened 2021-05-18
fix(Relation): remove relation

BotellaA:
I also removed the deprecation since we are not really sure about the thresold value yet.

### DIFF
--- a/src/geode/model/mixin/core/relationships.cpp
+++ b/src/geode/model/mixin/core/relationships.cpp
@@ -180,6 +180,16 @@ namespace geode
             return do_add_relation( from, to );
         }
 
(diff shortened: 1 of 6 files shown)


src/geode/model/mixin/core/relationships.cpp
```diff
@@ -180,6 +180,16 @@ namespace geode
             return do_add_relation( from, to );
         }
 
+        void remove_relation( const uuid& id1, const uuid& id2 )
+        {
+            if( const auto relation = relation_index( id1, id2 ) )
```
panquez, line 185: relation is a keyword on windows, so is it a good variable name?

panquez (approved):
Except variable name question, LGTM

PR #323, opened 2021-06-21
fix(Projection): missing point/plane projection

BotellaA:


include/geode/model/mixin/core/vertex_identifier.h
```diff
@@ -198,3 +198,17 @@ namespace geode
...
     };
 } // namespace geode
+
+namespace std
+{
+    template <>
+    struct hash< geode::MeshComponentVertex >
+    {
+    public:
+        size_t operator()( const geode::MeshComponentVertex& mcv ) const
```
panquez, line 208: ?

panquez (approved):

PR #325, opened 2021-08-13
fix(VertexCycle): expose sort only function

BotellaA:


panquez (approved):
Is polygon_sort the right name ? What About a VertexCycle of 2 elements ?

BotellaA:
> Is polygon_sort the right name ? What About a VertexCycle of 2 elements ?

It is not necessary for 2 elements. I really meant polygon and not VertexCycle.

panquez:
> > Is polygon_sort the right name ? What About a VertexCycle of 2 elements ?
> 
> It is not necessary for 2 elements. I really meant polygon and not VertexCycle.

I really do not understand the difference betwwen polygon and VertexCycle. If I don't I suppose other people may not understand.

PR #340, opened 2021-10-01
feat(Model): helpers to convert model into mesh

BotellaA:


BotellaA:
I think there is a smarter way to create vertices using unique vertices order. So output vertex order could be linked to unique vertex of the model on some cases.

Need to modify this PR

PR #355, opened 2021-12-02
fix(SurfaceMesh): add safety check while computing polygons around vertex

BotellaA:


src/geode/mesh/core/surface_mesh.cpp
```diff
@@ -114,6 +114,63 @@ namespace
...
+                    "[SurfaceMesh::polygons_around_vertex] Wrong polygon "
+                    "around vertex" );
+                polygons.push_back( next_edge );
+                cur_polygon_edge = mesh.polygon_adjacent_edge( next_edge );
+                safety_count++;
+            }
+        }
+        OPENGEODE_EXCEPTION( safety_count < MAX_SAFETY_COUNT,
+            "[SurfaceMesh::polygons_around_vertex] Too many polygons around "
+            "vertex ",
```
panquez, line 169: Maybe the message should tell that it is maybe a symptom for a bug

panquez (approved):

PR #356, opened 2021-12-22
feat(RegularGrid): Added support for vertices and their attributes on the regular grid.

MelchiorSchuh:


include/geode/mesh/core/regular_grid.h
```diff
@@ -93,6 +93,20 @@ namespace geode
         absl::optional< Index > previous_cell(
             const Index& index, index_t direction ) const;
 
+        index_t nb_vertices() const;
+
+        index_t nb_vertices( index_t direction ) const;
+
+        Index vertex_index( index_t index ) const;
```
BotellaA, line 100: Index represents both VertexIndex and CellIndex. Should we explicily separate them ? Maybe with two structs ? This will be a BREAKING change..

@panquez what do you think ?

MelchiorSchuh: True, but any CellIndex is the same as the VertexIndex of its corner, so I think it also makes sense to keep the same struct for the two (as was already the case with the 'Point<dim> point(const Index& index) const' function).

BotellaA: Yes but it does not refer to the same concept. Even if their implementations are equal. I think this is the same than PolygonVertex and PolygonEdge.

src/geode/mesh/core/regular_grid.cpp
```diff
@@ -132,6 +143,73 @@ namespace geode
...
+        {
+            return cells_number_.at( direction ) + 1;
+        }
+
+        index_t vertex_index( const Index& index ) const
+        {
+            index_t vertex_id{ 0 };
+            for( const auto d : LRange{ dimension } )
+            {
+                OPENGEODE_ASSERT( index[d] < cells_number_[d] + 1,
```
BotellaA, line 156: I do not really like the use of cells_number + 1 everywhere. Maybe use your new nb_vertices methods to avoid this implementation detail in the entire class.

src/geode/mesh/core/regular_grid.cpp
```diff
@@ -132,6 +143,73 @@ namespace geode
...
+                vertex_id[dimension - d - 1] = value;
+                index -= value * offset;
+            }
+            return vertex_id;
+        }
+
+        absl::optional< Index > next_vertex(
+            const Index& index, index_t direction ) const
+        {
+            if( index[direction] + 1 < nb_cells( direction ) )
```
BotellaA, line 192: nb_cells or nb_vertices ?

src/geode/mesh/core/regular_grid.cpp
```diff
@@ -132,6 +143,73 @@ namespace geode
...
+        index_t vertex_index( const Index& index ) const
+        {
+            index_t vertex_id{ 0 };
+            for( const auto d : LRange{ dimension } )
+            {
+                OPENGEODE_ASSERT( index[d] < nb_vertices( d ),
+                    "[RegularGrid::vertex_index] Invalid index" );
+
+                index_t offset{ 1 };
+                for( const auto d2 : Range{ d } )
```
BotellaA, line 160:
```suggestion
                for( const auto d2 : LRange{ d } )
```

src/geode/mesh/core/regular_grid.cpp
```diff
@@ -132,6 +143,73 @@ namespace geode
...
+
+        Index vertex_index( index_t index ) const
+        {
+            OPENGEODE_ASSERT( index < vertex_attribute_manager().nb_elements(),
+                "[RegularGrid::vertex_index] Invalid index" );
+            Index vertex_id;
+            for( const auto d : LRange{ dimension } )
+            {
+                index_t offset{ 1 };
+                for( const auto d2 : Range{ dimension - d - 1 } )
```
BotellaA, line 177:
```suggestion
                for( const auto d2 : LRange{ dimension - d - 1 } )
```

BotellaA (commented):
Minor changes. We need to decide if we need specific Index classes or not before merging this.

include/geode/mesh/core/regular_grid.h
panquez: I agree with this last proposal from Arnaud. IMo CellIndex and VertexIndex should be different

BotellaA (approved):

PR #382, opened 2022-02-12
fix(ProgressLogger): display starting message and add duration in final message

BotellaA:
What do you think about this?

### DIFF
--- a/src/geode/basic/progress_logger.cpp
+++ b/src/geode/basic/progress_logger.cpp
@@ -23,6 +23,7 @@
 
 #include <geode/basic/progress_logger.h>
 
+#include <absl/synchronization/mutex.h>
 #include <absl/time/clock.h>
 #include <absl/time/time.h>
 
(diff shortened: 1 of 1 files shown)


src/geode/basic/progress_logger.cpp
```diff
@@ -42,36 +43,42 @@ namespace geode
...
         {
+            Logger::info( message_, " starts" );
         }
 
         ~Impl()
         {
             if( current_ == max_number_ )
             {
-                Logger::info( message_, " complete" );
+                Logger::info( message_, " complete in ",
```
panquez, line 56: What is the output of absl::FormatDuration? Is it only a number or does it return also time unit (s, min or h)?

BotellaA: It depends on the duration, for duration < to 1s it displays ms or ns, for more it displays h,m,s when possible. Here is an example of our test
[2022-02-12 14:47:37.230] [console] [info] Cool message complete in 4.336392777s

BotellaA:
@MelchiorSchuh not sure about the "starts" and "complete"... Is it correct English?

BotellaA:
Maybe the start message is not needed if we output the time duration in the complete message?

MelchiorSchuh:
I'd say "completed in" instead of "complete in".
For the start message, it has some sense, to have start - progress - end and to make sure the process actually started.

PR #394, opened 2022-03-07
fix(RegularGrid): removed deprecated functions, cleared multiple defi…

MelchiorSchuh:
…nitions, clarified grid accessors names.


include/geode/mesh/core/regular_grid.h
```diff
@@ -101,59 +98,65 @@ namespace geode
 
         const Point< dimension >& origin() const;
 
+        AttributeManager& cell_attribute_manager() const;
```
BotellaA, line 101: Why moving them? Most of the classes attribute managers are at the bottom. But why not.

src/geode/mesh/core/regular_grid.cpp
```diff
@@ -346,17 +338,12 @@ namespace geode
...
-        }
-
-        absl::optional< GridCellIndices< dimension > > cell(
+        absl::optional< GridCellsAroundVertex< dimension > > cell(
             const Point< dimension >& query ) const
         {
-            GridCellIndices< dimension > indices;
-            GridCellIndex< dimension > min;
-            GridCellIndex< dimension > max;
+            GridCellsAroundVertex< dimension > cells_around_point;
```
BotellaA, line 344: To declare only at line 370?

BotellaA (approved):

include/geode/mesh/core/regular_grid.h
MelchiorSchuh: I just felt that theses accessors to important features should be visible before the functions computing some stuff on the class.

panquez (approved):

PR #412, opened 2022-03-29
fix(Meshes): Passing the copy() function as public without a Key, but…

MelchiorSchuh:
… with an Exception to make sure the mesh is not initialized before the copy is made.

### DIFF
--- a/src/geode/mesh/builder/tetrahedral_solid_builder.cpp
+++ b/src/geode/mesh/builder/tetrahedral_solid_builder.cpp
@@ -111,6 +111,10 @@ namespace geode
     void TetrahedralSolidBuilder< dimension >::copy(
         const TetrahedralSolid< dimension >& tetrahedral_solid )
     {
+        OPENGEODE_EXCEPTION( tetrahedral_solid_.nb_vertices() == 0
+                                 && tetrahedral_solid_.nb_polyhedra() == 0,
+            "[TetrahedralSolidBuilder::copy] Cannot copy a mesh into an "
+            "already initialized mesh." );
         SolidMeshBuilder< dimension >::copy( tetrahedral_solid );
     }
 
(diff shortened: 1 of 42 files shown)


BotellaA (approved):

src/geode/mesh/builder/tetrahedral_solid_builder.cpp
```diff
@@ -111,6 +111,11 @@ namespace geode
     void TetrahedralSolidBuilder< dimension >::copy(
         const TetrahedralSolid< dimension >& tetrahedral_solid )
     {
+        OPENGEODE_EXCEPTION( tetrahedral_solid_.nb_vertices() == 0
+                                 && tetrahedral_solid_.nb_polyhedra() == 0,
+            "[TetrahedralSolidBuilder::copy] Cannot copy a mesh into an "
+            "already "
+            "initialized mesh." );
```
panquez, line 118:
```suggestion
            "[TetrahedralSolidBuilder::copy] Cannot copy a mesh into an "
            "already initialized mesh." );
```

src/geode/mesh/builder/triangulated_surface_builder.cpp
```diff
@@ -111,6 +111,11 @@ namespace geode
     void TriangulatedSurfaceBuilder< dimension >::copy(
         const TriangulatedSurface< dimension >& triangulated_surface )
     {
+        OPENGEODE_EXCEPTION( triangulated_surface_.nb_vertices() == 0
+                                 && triangulated_surface_.nb_polygons() == 0,
+            "[TriangulatedSurfaceBuilder::copy] Cannot copy a mesh into an "
+            "already "
+            "initialized mesh." );
```
panquez, line 118:
```suggestion
            "[TriangulatedSurfaceBuilder::copy] Cannot copy a mesh into an "
            "already initialized mesh." );

```

panquez (approved):

panquez (approved):

PR #415, opened 2022-04-01
fix(Basic): allowed to map a type to another type in GenericMapping

panquez:


MelchiorSchuh (approved):

include/geode/basic/mapping.h
```diff
@@ -70,11 +70,14 @@ namespace geode
         absl::flat_hash_map< T2, T1 > out2in_;
     };
 
-    template < typename T, index_t Capacity >
+    template < typename T >
+    using Storage = absl::InlinedVector< T, 1 >;
```
BotellaA, line 74: Why not nesting it inside GenericMapping?

PR #419, opened 2022-04-15
CachedValue

BotellaA:


MelchiorSchuh (approved):

BotellaA:
I added first implementation of PolygonsAroundVertex. Let's first review this.

include/geode/mesh/core/surface_mesh.h
```diff
@@ -328,7 +328,8 @@ namespace geode
          * @param[in] vertex_id Index of the vertex.
          * @pre This function needs that polygon adjacencies are computed
          */
-        PolygonsAroundVertex polygons_around_vertex( index_t vertex_id ) const;
+        const PolygonsAroundVertex& polygons_around_vertex(
```
BotellaA, line 331: Should we change this? This is not a breaking change.

panquez, line 331: You should check when it is used to be sure we use correctly the ref

panquez (approved):

include/geode/mesh/core/surface_mesh.h
BotellaA: I did it in OG. By default it is fine because it was a return by value, so worst case we do a copy instead of a ref.

src/geode/mesh/core/surface_mesh.cpp
```diff
@@ -249,6 +257,31 @@ namespace geode
...
+        void reset_polygons_around_vertex( index_t vertex_id )
+        {
+            polygons_around_vertex_->modify_value(
+                vertex_id, []( CachedPolygons& value ) { value.reset(); } );
+        }
+
+        const PolygonsAroundVertex& polygons_around_vertex(
+            const SurfaceMesh< dimension >& mesh,
+            index_t vertex_id,
+            absl::optional< PolygonVertex > first_polygon ) const
```
MelchiorSchuh, line 269: Why not put the mesh as an attribute of the impl class ? In order for the mesh and impl class to be owners of their attributes ? ?

src/geode/mesh/builder/surface_mesh_builder.cpp
```diff
@@ -366,6 +395,7 @@ namespace geode
                         polygons_around[1], polygon_vertex_id );
                 }
             }
+            reset_polygons_around_vertex( polygon_vertex_id );
```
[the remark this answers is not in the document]
BotellaA: Do not confuse polygon_around_vertex() with polygon_vertex(), here we want to reset the value because we are changing the polygon vertex of the polygon

src/geode/mesh/core/solid_mesh.cpp
```diff
@@ -456,26 +507,66 @@ namespace geode
...
+        }
+
+        const detail::PolyhedraAroundVertexImpl& update_polyhedra_around_vertex(
+            const SolidMesh< dimension >& mesh,
+            const index_t vertex_id,
+            const absl::optional< PolyhedronVertex >& first_polyhedron ) const
+        {
+            const auto& cached = polyhedra_around_vertex_->value( vertex_id );
+            const auto& polyhedra = cached.value().polyhedra;
+            if( !cached.computed()
```
BotellaA, line 527: Main question is here: when to (re)compute the value?

MelchiorSchuh: I'd say the part checking if it was already computed belongs in "polyhedra_around_vertex", and update_polyhedra_around_vertex, as its name states, systematically updates it.

PR #420, opened 2022-04-20
fix(SolidMesh): add method to know if a vertex is on border or not

panquez:


MelchiorSchuh (approved):
Add it to tests (and python bindings) ?

BotellaA (approved):

PR #437, opened 2022-05-09
fix(RegularGrid): Added tests to remove the possibility to create a g…

MelchiorSchuh:
…rid with no cells in a direction or with a cell length smaller than epsilon

### DIFF
--- a/src/geode/mesh/core/regular_grid.cpp
+++ b/src/geode/mesh/core/regular_grid.cpp
@@ -59,20 +59,30 @@ namespace geode
               cells_number_( std::move( cells_number ) ),
               cells_length_( std::move( cells_length ) )
         {
-            double nb_cells_double{ 1 };
+            for( const auto direction : LRange{ dimension } )
+            {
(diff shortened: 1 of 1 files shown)


src/geode/mesh/core/regular_grid.cpp
```diff
@@ -59,20 +59,30 @@ namespace geode
               cells_number_( std::move( cells_number ) ),
               cells_length_( std::move( cells_length ) )
         {
-            double nb_cells_double{ 1 };
+            for( const auto direction : LRange{ dimension } )
+            {
+                OPENGEODE_EXCEPTION( cells_length_[direction] < global_epsilon,
```
BotellaA, line 64:
```suggestion
                OPENGEODE_EXCEPTION( cells_length_[direction] > global_epsilon,

```

BotellaA (changes requested):

panquez (approved):

BotellaA (approved):

PR #443, opened 2022-05-13
fix(Mesh): add more around methods

panquez:


src/geode/mesh/core/tetrahedral_solid.cpp
```diff
@@ -161,6 +161,41 @@ namespace geode
         return result;
     }
 
+    template < index_t dimension >
+    PolyhedraAroundEdge TetrahedralSolid< dimension >::polyhedra_around_edge(
```
BotellaA, line 165: Do we really need to reimplement it?

src/geode/mesh/core/tetrahedral_solid.cpp
```diff
@@ -161,6 +161,41 @@ namespace geode
...
+    }
+
+    template < index_t dimension >
+    PolyhedraAroundEdge TetrahedralSolid< dimension >::polyhedra_around_edge(
+        const std::array< index_t, 2 >& vertices,
+        index_t first_polyhedron ) const
+    {
+        PolyhedraAroundEdge result{ first_polyhedron };
+        const auto v0 =
+            this->vertex_in_polyhedron( first_polyhedron, vertices[0] ).value();
```
BotellaA, line 180: What if the vertex is not in the polyhedron? Maybe add a safety

BotellaA (approved):

PR #456, opened 2022-05-25
fix(Geometry): implement an API function to get lexicographic mapping…

panquez:
… of a list of points

### DIFF
--- a/include/geode/geometry/morton.h
+++ b/include/geode/geometry/morton.h
@@ -22,5 +22,7 @@
  */
 
 #pragma once
+#pragma message(                                                               \
+    "#include <geode/geometry/morton.h> is deprecated. Use e.g. #include <geode/geometry/points_sort>" )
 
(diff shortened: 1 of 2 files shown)


MelchiorSchuh (approved):

include/geode/geometry/morton.h
```diff
@@ -23,19 +23,4 @@
...
-    FORWARD_DECLARATION_DIMENSION_CLASS( Point );
-} // namespace geode
-
-namespace geode
-{
-    template < index_t dimension >
-    std::vector< index_t > morton_sort(
-        absl::Span< const Point< dimension > > points );
-} // namespace geode
+#include <geode/geometry/points_sort.h>
```
BotellaA, line 26: Should we add a deprecation message like in basic object ?

BotellaA (approved):

PR #458, opened 2022-05-30
fix(VertexCycle): Added a version of the VertexCycle which keeps the …

MelchiorSchuh:
…orientation of the vertices.

### DIFF
--- a/include/geode/mesh/core/detail/vertex_cycle.h
+++ b/include/geode/mesh/core/detail/vertex_cycle.h
@@ -52,21 +52,17 @@ namespace geode
         }
 
         template < typename Container >
-        inline void rotate( Container& vertices )
+        inline void oriented_rotate( Container& vertices )
         {
             if( vertices.size() < 2 )
             {
                 return;
             }
             polygon_sort( vertices );
-            if( vertices[1] > vertices.back() )
-            {
-                std::reverse( vertices.begin() + 1, vertices.end() );
-            }
         }
 
         template <>
-        inline void rotate( std::array< index_t, 2 >& vertices )
+        inline void oriented_rotate( std::array< index_t, 2 >& vertices )
         {
             if( vertices.front() > vertices.back() )
             {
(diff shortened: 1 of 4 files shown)


tests/mesh/test-vertex-cycle.cpp
```diff
@@ -66,9 +66,42 @@ void test_vertex_cycle()
...
+    const geode::detail::OrientedVertexCycle< std::array< geode::index_t, 2 > >
+        cycle7{ { 3, 1 } };
+
+    OPENGEODE_EXCEPTION( cycle1 == cycle2,
+        "[Test] Wrong result for operator== with cycle1 and cycle2" );
+    OPENGEODE_EXCEPTION( cycle1 != cycle3,
+        "[Test] Wrong result for operator== with cycle1 and cycle3" );
+    OPENGEODE_EXCEPTION( cycle1 != cycle4,
+        "[Test] Wrong result for operator== with cycle1 and cycle4" );
+    OPENGEODE_EXCEPTION( cycle1.is_opposite( cycle4 ),
```
panquez, line 93: check message

include/geode/mesh/core/detail/vertex_cycle.h
```diff
@@ -74,6 +74,22 @@ namespace geode
...
+            }
+            polygon_sort( vertices );
+        }
+
+        template <>
+        inline void oriented_rotate( std::array< index_t, 2 >& vertices )
+        {
+            rotate( vertices );
+        }
+
```
panquez, line 92: is it different from l.68 ? should we have two implementations or use the same?

panquez (approved):

include/geode/mesh/core/detail/vertex_cycle.h
```diff
@@ -52,80 +52,147 @@ namespace geode
...
 
+        template < typename Container >
+        inline void rotate( Container& vertices )
+        {
+            oriented_rotate( vertices );
+            remove_orientation( vertices );
+        }
+
         template <>
         inline void rotate( std::array< index_t, 2 >& vertices )
```
BotellaA, line 90: Is `rotate` still used?

BotellaA (approved):

include/geode/mesh/core/detail/vertex_cycle.h
MelchiorSchuh: No, but wouldn't removing it be a breaking change ?

PR #461, opened 2022-05-31
Fix/polyhedron facets from unique vertices facet

MelchiorSchuh:


BotellaA (approved):

include/geode/model/helpers/component_mesh_queries.h
```diff
@@ -38,8 +38,8 @@ namespace geode
 
 namespace geode
 {
-    PolyhedraAroundFacet block_mesh_polyhedra_from_unique_vertices_facet(
-        const BRep& model,
-        const Block3D& block,
-        absl::Span< const index_t > facet_unique_vertices );
+    newPolyhedraAroundFacet opengeode_model_api
```
panquez, line 41: This is breaking change. Could you leave the function as it was and create a `new_xxxx` one?

panquez (approved):

PR #465, opened 2022-06-06
feat(BRepQueries): Added queries to access the facet(s) of a block co…

MelchiorSchuh:
…rresponding to the polygon of a surface of a BRep.

### DIFF
--- a/include/geode/model/helpers/component_mesh_queries.h
+++ b/include/geode/model/helpers/component_mesh_queries.h
@@ -26,11 +26,14 @@
 #include <memory>
 
 #include <geode/mesh/core/solid_mesh.h>
+#include <geode/mesh/core/surface_mesh.h>
 
 #include <geode/model/common.h>
 
 namespace geode
 {
+    FORWARD_DECLARATION_DIMENSION_CLASS( Surface );
+    ALIAS_3D( Surface );
     FORWARD_DECLARATION_DIMENSION_CLASS( Block );
     ALIAS_3D( Block );
     class BRep;
@@ -38,8 +41,44 @@ namespace geode
 
 namespace geode
 {
+    struct BlockPolyhedraFacetVertices
+    {
+        index_t nb_facets() const
+        {
+            index_t counter{ 0 };
+            if( oriented_polyhedron_vertices )
+            {
+                counter++;
+            }
+            if( opposite_polyhedron_vertices )
+            {
+                counter++;
+            }
+            return counter;
+        }
+
+        absl::optional< PolygonVertices > oriented_polyhedron_vertices;
+        absl::optional< PolygonVertices > opposite_polyhedron_vertices;
+    };
+
+    PolygonVertices opengeode_model_api surface_polygon_unique_vertices(
+        const BRep& model, const Surface3D& surface, index_t polygon_id );
+
     newPolyhedraAroundFacet opengeode_model_api
-        block_mesh_polyhedra_from_unique_vertices_facet( const BRep& model,
+        block_mesh_polyhedra_from_surface_polygon( const BRep& model,
             const Block3D& block,
-            absl::Span< const index_t > facet_unique_vertices );
+            const Surface3D& surface,
+            index_t polygon_id );
+
+    absl::InlinedVector< PolygonVertices, 2 >
+        block_vertices_from_surface_polygon( const BRep& model,
+            const Block3D& block,
+            const Surface3D& surface,
+            index_t polygon_id );
+
+    BlockPolyhedraFacetVertices oriented_block_vertices_from_surface_polygon(
+        const BRep& model,
+        const Block3D& block,
+        const Surface3D& surface,
+        index_t polygon_id );
 } // namespace geode
--- a/src/geode/model/helpers/component_mesh_queries.cpp
+++ b/src/geode/model/helpers/component_mesh_queries.cpp
@@ -25,7 +25,10 @@
 
 #include <absl/container/inlined_vector.h>
 
+#include <geode/mesh/core/detail/vertex_cycle.h>
+
 #include <geode/model/mixin/core/block.h>
+#include <geode/model/mixin/core/surface.h>
 #include <geode/model/representation/core/brep.h>
 
 namespace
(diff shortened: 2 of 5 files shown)


src/geode/model/helpers/component_mesh_queries.cpp
```diff
@@ -137,4 +217,77 @@ namespace geode
             "triangles and quadrangles." );
         return {};
     }
+
+    absl::InlinedVector< PolygonVertices, 2 >
+        block_vertices_from_surface_polygon( const BRep& model,
```
BotellaA, line 222: We need to discuss what this function does. Looks too much according to its name.

src/geode/model/helpers/component_mesh_queries.cpp
```diff
@@ -109,10 +112,87 @@ namespace
...
+        {
+            const auto block_vertex_indices = model.mesh_component_vertices(
+                polygon_unique_vertices[polygon_vertex], block.id() );
+            if( block_vertex_indices.size() == 1 )
+            {
+                block_facet_vertices.emplace_back( block_vertex_indices[0] );
+            }
+            else
+            {
+                return absl::nullopt;
```
panquez, line 134: So it does no return something if surface is internal?

src/geode/model/helpers/component_mesh_queries.cpp
```diff
@@ -137,4 +217,77 @@ namespace geode
...
+                    facets_block_vertices.back() );
+                std::swap( facets_unique_vertices.front(),
+                    facets_unique_vertices.back() );
+            }
+            else
+            {
+                OPENGEODE_ASSERT( polygon_unique_vertices_cycle.is_opposite(
+                                      { facets_unique_vertices[1] } ),
+                    "[block_vertices_from_surface_polygon] The block facets "
+                    "found from the polygon vertices have the same normal." );
```
panquez, line 266:
```suggestion
                    "found from the polygon vertices have the same orientation." );
```

src/geode/model/helpers/component_mesh_queries.cpp
panquez: and maybe the function is a bit too long

src/geode/model/helpers/component_mesh_queries.cpp
MelchiorSchuh: If the surface is internal or if the a polygon vertex has several mesh_component_vertices associated, it returns a null_opt (e.g. it cannot find the vertices from this information only).

include/geode/model/helpers/component_mesh_queries.h
```diff
@@ -26,20 +26,65 @@
...
 
 namespace geode
 {
+    FORWARD_DECLARATION_DIMENSION_CLASS( Surface );
+    ALIAS_3D( Surface );
     FORWARD_DECLARATION_DIMENSION_CLASS( Block );
     ALIAS_3D( Block );
     class BRep;
+
+    struct BlockPolyhedronsFacetVertices
```
BotellaA, line 41: We used to declare this not in the namespace for forward declarations.

src/geode/model/helpers/component_mesh_queries.cpp
```diff
@@ -137,4 +162,177 @@ namespace geode
...
+            "is neither boundary nor internal to the given block in the given "
+            "model." );
+        const auto polygon_unique_vertices =
+            surface_polygon_unique_vertices( model, surface, polygon_id );
+        BlockPolyhedronsFacetVertices polygon_to_block_vertices;
+        if( auto block_vertices =
+                block_unique_facet_vertices_from_surface_polygon(
+                    model, block, polygon_unique_vertices ) )
+        {
+            polygon_to_block_vertices.oriented_polyhedron_vertices =
```
BotellaA, line 297: Why oriented by default ?

include/geode/model/helpers/component_mesh_queries.h
panquez, line 41:
```suggestion
    struct BlockPolyhedraFacetVertices
```
?

src/geode/model/helpers/component_mesh_queries.cpp
```diff
@@ -137,4 +162,234 @@ namespace geode
...
+            "is neither boundary nor internal to the given block in the given "
+            "model." );
+        const auto polygon_unique_vertices =
+            surface_polygon_unique_vertices( model, surface, polygon_id );
+        BlockPolyhedraFacetVertices p_to_block_vertices;
+        if( auto block_vertices =
+                block_unique_facet_vertices_from_surface_polygon(
+                    model, block, polygon_unique_vertices ) )
+        {
+            p_to_block_vertices.oriented_polyhedron_vertices =
```
BotellaA, line 354: Why still oriented by default? Why not have the block_mesh_polyhedra_from_surface_polygon return a struct telling you which facet is oriented or not?

BotellaA, line 354: Why still oriented by default? Why not have the block_mesh_polyhedra_from_surface_polygon return a struct telling you which facet is oriented or not?

include/geode/model/helpers/component_mesh_queries.h
```diff
@@ -26,20 +26,68 @@
...
+                counter++;
+            }
+            if( opposite_polyhedron_vertices )
+            {
+                counter++;
+            }
+            return counter;
+        }
+
+        absl::optional< PolygonVertices > any_of() const
```
panquez, line 60: Not used anymore?

src/geode/model/helpers/component_mesh_queries.cpp
```diff
@@ -137,4 +162,234 @@ namespace geode
...
+    }
+
+    /*!
+     * Returns the result of polygon_vertices_to_block_facets_vertices, but with
+     * the lists oriented so that the first PolygonVertices of each vector is
+     * linked to the facet with the same orientation as the given vertices.
+     */
+    std::tuple< absl::InlinedVector< geode::PolygonVertices, 2 >,
+        absl::InlinedVector< geode::PolygonVertices, 2 >,
+        bool >
```
panquez, line 209: maybe i am wrong, but why do not you use `BlockPolyhedraFacetVertices` as return type?

panquez (approved):
I have a bit of difficulty to understand all the cpp code, so my comments are maybe wrong

src/geode/model/helpers/component_mesh_queries.cpp
MelchiorSchuh: Yes, they're not the same object, this function returns two lists: the PolygonVertices representing the vertices in the block of each facet, and the PolygonVertices rerpesenting the unique vertices associated (with a corresponding numbering and orientation in the two lists.

BotellaA (approved):

PR #474, opened 2022-06-16
Fix/filtermodel

panquez:


include/geode/model/helpers/simplicial_brep_creator.h
```diff
@@ -55,9 +55,18 @@ namespace geode
         std::vector< uuid > create_surfaces( absl::Span< const uuid > lines,
             absl::Span< const SurfaceDefinition > definitions );
 
+        std::vector< uuid > create_surfaces( absl::Span< const uuid > corners,
+            absl::Span< const uuid > lines,
+            absl::Span< const SurfaceDefinition > definitions );
```
BotellaA, line 60: Should we keep both? Maybe yes for now but look weird. We might need to rethink the class flow (someday)

src/geode/model/helpers/model_component_filter.cpp
```diff
@@ -0,0 +1,110 @@
...
+        std::vector< geode::uuid > surfaces_to_delete;
+        surfaces_to_delete.reserve( model.nb_surfaces() );
+        for( const auto& surface : model.surfaces() )
+        {
+            if( model.nb_incidences( surface.id() ) == 0
+                && model.nb_embeddings( surface.id() ) == 0 )
+            {
+                surfaces_to_delete.push_back( surface.id() );
+            }
+        }
```
BotellaA, line 48: You could even extract this code by taking a template Range in parameters and the number of elements to reserve and returning the vector to delete.

PR #493, opened 2022-06-29
fix(Input): return the loaded object

BotellaA:
BREAKING CHANGE: signature of the read method changed

### DIFF
--- a/include/geode/mesh/core/regular_grid.h
+++ b/include/geode/mesh/core/regular_grid.h
@@ -67,6 +67,8 @@ namespace geode
         friend class bitsery::Access;
 
     public:
+        RegularGrid();
+
         RegularGrid( Point< dimension > origin,
             std::array< index_t, dimension > cells_number,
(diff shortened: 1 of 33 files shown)


include/geode/mesh/core/regular_grid.h
```diff
@@ -67,6 +67,8 @@ namespace geode
         friend class bitsery::Access;
 
     public:
+        RegularGrid();
```
MelchiorSchuh, line 70: How do you define the creation of the grid with no parameters ?

MelchiorSchuh (approved):

include/geode/mesh/core/regular_grid.h
BotellaA: You don't. This is wrong at the moment but it simplifies SO MUCH this PR. 
In v11, we will need this constructor to use the Grid as a Block mesh. Then, we will add more methods to the builder to create the grid later with some parameters.

PR #497, opened 2022-07-01
fix(ModelHelpers): cut surfaces along internal lines

BotellaA:


src/geode/model/helpers/cut_along_internal_lines.cpp
```diff
@@ -0,0 +1,274 @@
...
+            }
+
+            MCVmappings duplicate_points( const Surface< dimension >& surface )
+            {
+                auto builder = builder_.surface_mesh_builder( surface.id() );
+                remove_adjacencies_along_internal_lines( surface, *builder );
+                return split_points( surface, *builder );
+            }
+
+            MCVmappings split_points( const Surface< dimension >& surface,
```
panquez, line 110: I will swap these two method names `duplicate_points` and `split_points`

panquez (approved):

PR #498, opened 2022-07-02
feat(String): add string manipulations

BotellaA:


panquez (approved):

MelchiorSchuh (approved):
I see the point, but I'm not sure if this really brings something more than the absl functions and if it doesn't over-specialize them (especially string_split).

BotellaA:
> I see the point, but I'm not sure if this really brings something more than the absl functions and if it doesn't over-specialize them (especially string_split).

It is important, especially for string_split. We are using it with different parameters in the entire codebase. Not by choice but by lack of knowledge on how to use it

PR #505, opened 2022-07-05
fix(Attribute): Added a function to change a VariableAttribute so it …

MelchiorSchuh:
…can be set as interpolable.

### DIFF
--- a/include/geode/basic/attribute.h
+++ b/include/geode/basic/attribute.h
@@ -128,6 +128,11 @@ namespace geode
         {
         }
 
+        void set_properties( AttributeProperties new_properties )
+        {
+            properties_ = std::move( new_properties );
+        }
+
     private:
         AttributeProperties properties_;
     };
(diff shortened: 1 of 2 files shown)


include/geode/basic/attribute.h
```diff
@@ -370,6 +375,11 @@ namespace geode
             set_value( to_element, interpolation.compute_value( *this ) );
         }
 
+        void set_to_interpolable()
```
BotellaA, line 378: Open question : 
void set_properties( AttributeProperties ) ?

panquez: Instead of `change_properties`

panquez (approved):

include/geode/basic/attribute.h
```diff
@@ -128,6 +128,11 @@ namespace geode
         {
         }
 
+        virtual void set_properties( AttributeProperties new_properties )
```
BotellaA, line 131: No virtual here, this is not an override of the implementation.

BotellaA (changes requested):

BotellaA (approved):

PR #508, opened 2022-07-08
fix(Geometry): better handle flat triangle area computation

BotellaA:


src/geode/geometry/signed_mensuration.cpp
```diff
@@ -34,32 +35,27 @@ namespace geode
...
+        }
+        return std::sqrt( area2 );
     }
 
     double triangle_signed_area(
         const Triangle3D& triangle, const Vector3D& direction )
     {
         const auto area = triangle_area( triangle );
-        try
+        if( const auto normal = triangle.new_normal() )
```
BotellaA, line 54: The try/catch was here before using the new_normal. It is no longer needed.

panquez (approved):

PR #514, opened 2022-07-18
better return info fom conversion between Section/BRep

BotellaA:


src/geode/model/helpers/convert_brep_section.cpp
```diff
@@ -37,25 +37,50 @@
...
-                brep, section, builder ) );
-        mappings.emplace( Line2D::component_type_static(),
-            detail::copy_line_components_without_type(
-                brep, section, builder ) );
-        mappings.emplace( Surface2D::component_type_static(),
-            detail::copy_surface_components_without_type(
-                brep, section, builder ) );
-        builder.copy_relationships( mappings, brep );
-
+        DEBUG_CONST auto mappings =
```
panquez, line 81: why DEBUG_CONST?

panquez (approved):

src/geode/model/helpers/convert_brep_section.cpp
panquez: I agree

PR #530, opened 2022-07-25
fix(Basic): Added a function to get the filepath without extension of…

MelchiorSchuh:
… a function.

### DIFF
--- a/src/geode/basic/filename.cpp
+++ b/src/geode/basic/filename.cpp
@@ -38,4 +38,10 @@ namespace geode
         ghc::filesystem::path filepath{ to_string( path ) };
         return filepath.filename().replace_extension( "" ).string();
     }
+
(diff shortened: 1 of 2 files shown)


src/geode/basic/filename.cpp
```diff
@@ -38,4 +38,10 @@ namespace geode
         ghc::filesystem::path filepath{ to_string( path ) };
         return filepath.filename().replace_extension( "" ).string();
     }
+
+    std::string filepath_without_extension( absl::string_view path )
+    {
+        ghc::filesystem::path filepath{ geode::to_string( path ) };
```
panquez, line 44:
```suggestion
        ghc::filesystem::path filepath{ to_string( path ) };
```

panquez (approved):

PR #531, opened 2022-07-28
feat(ProgressLogger): external client support

BotellaA:


panquez (commented):
Peux-tu faire un test qui vérifie que ça marche bien ?

BotellaA:
There is a progress logger test. It is mainly graphical but it works.

panquez (approved):

PR #532, opened 2022-08-04
fix(ProgressLogger): add a Manager to better handle extensibility

BotellaA:
There is an example in the test


include/geode/basic/progress_logger_manager.h
```diff
@@ -0,0 +1,60 @@
...
+        static void register_client(
+            std::unique_ptr< ProgressLoggerClient >&& client );
+
+        static void start( const std::string& message, index_t nb_steps );
+
+        static void update( index_t current_step, index_t nb_steps );
+
+        static void complete();
+
+        static void failed();
```
BotellaA, line 50: Maybe these could be protected by a key so only the ProgressLogger can call them.

include/geode/basic/console_progress_logger_client.h
```diff
@@ -35,15 +35,17 @@ namespace geode
...
     private:
-        void start( index_t nb_steps ) override;
+        void start( const std::string& message, index_t nb_steps ) override;
 
         void update( index_t current_step, index_t nb_steps ) override;
 
-        void end( index_t current_step, index_t nb_steps ) override;
+        void complete() override;
+
+        void failed() override;
```
panquez, line 48: What are the purpose of these two methods? Is it intentional that one use "ed" but not the other one?

BotellaA: The goal is to signal the end of the progress and how it ended.

BotellaA: There is probably a better choice of words 😜

panquez: Ok i would say `completed()` and `failed()`

BotellaA: Could you do the changes please?

PR #535, opened 2022-09-03
feat(Database): new databse for geode objects

BotellaA:


include/geode/basic/database.h
```diff
@@ -0,0 +1,112 @@
...
+#include <geode/basic/identifier.h>
+#include <geode/basic/uuid.h>
+
+namespace geode
+{
+    class Identifier;
+    struct uuid;
+} // namespace geode
+
+namespace geode
```
BotellaA, line 39: I am not happy about almost all the names in this file...

src/geode/basic/database.cpp
```diff
@@ -0,0 +1,345 @@
...
+#include <absl/synchronization/mutex.h>
+#include <absl/time/time.h>
+
+#include <geode/basic/identifier.h>
+#include <geode/basic/pimpl_impl.h>
+#include <geode/basic/uuid.h>
+
+namespace
+{
+    constexpr auto DATA_EXPIRATION = absl::Seconds( 10 );
```
BotellaA, line 39: To discuss, 10s are for testing. A production value could be 5min?

src/geode/basic/database.cpp
```diff
@@ -0,0 +1,345 @@
...
+        absl::Mutex lock_;
+    };
+
+    class Database::Impl
+    {
+    public:
+        Impl( absl::string_view directory )
+            : directory_{ to_string( directory ) }
+        {
+            if( !ghc::filesystem::exists( directory_ ) )
```
BotellaA, line 145: What if the folder already exists? Should we still write in it?

src/geode/basic/database.cpp
```diff
@@ -0,0 +1,345 @@
...
+            const auto filename =
+                absl::StrCat( directory_, "/", data->id().string() );
+            std::ofstream file{ filename, std::ofstream::binary };
+            TContext context;
+            for( const auto& serializer : serializers_ )
+            {
+                serializer( std::get< 0 >( context ) );
+            }
+            Serializer archive{ context, file };
+            archive.ext( data, bitsery::ext::StdSmartPtr{} );
```
BotellaA, line 236: This can only be done for Bitsery-compatible objects. I think we could add this as mandatory. We do this database for internal usage. If yes, some of our objects need better bitsery support like BRep but it is not a problem

src/geode/basic/database.cpp
```diff
@@ -0,0 +1,345 @@
...
+        {
+            storage_->new_data_reference();
+        }
+
+        ~Impl()
+        {
+            storage_->delete_data_reference();
+        }
+
+        const Identifier& data() const
```
BotellaA, line 287: There is a potential bug for "basic" dev. 

If you take the reference 'A,' pass it to another class, and then destroy Data (which was temporarily holding 'A'). This is a lifetime issue for the reference because destroying the Data will trigger the delete_data_reference which implies this data is no longer used. So it could be released from memory by the database at some point. It does not matter if the reference was given to another class.

A way to avoid people to use the design wrong is to force the dev to manually call delete_data_reference. But I find this solution very verbose and constraining... Know your C++ better and add a bug warning in the doc?

tests/basic/test-database.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+struct Foo : public geode::Identifier
+{
+    Foo() = default;
+    Foo( double value ) : value_( value ) {}
+
+    template < typename Archive >
+    void serialize( Archive& archive )
+    {
+        archive.ext( *this, bitsery::ext::BaseClass< geode::Identifier >{} );
```
BotellaA, line 40: This is mandatory for serializing

tests/basic/test-database.cpp
```diff
@@ -0,0 +1,121 @@
...
+    context
+        .registerSingleBaseBranch< geode::Serializer, geode::Identifier, Foo >(
+            "Foo" );
+}
+
+void register_foo_deserializer( geode::PContext& context )
+{
+    context.registerSingleBaseBranch< geode::Deserializer, geode::Identifier,
+        Foo >( "Foo" );
+}
```
BotellaA, line 58: This is mandatory to (de)serialize directly from a smart pointer of an Identifier and let Bitsery saves/loads the right polymorphic type.

include/geode/basic/database.h
```diff
@@ -0,0 +1,112 @@
...
+                "[Database::register_data] Data is not a subclass of "
+                "Identifier" );
+            return register_unique_data( std::move( data ) );
+        }
+
+        void delete_data( const uuid& id );
+
+        Data get_data( const uuid& id ) const;
+
+        std::unique_ptr< Identifier > take_data( const uuid& id );
```
BotellaA, line 100: We could template this method BUT we need to make sure to handle cast failure properly. What behavior if a cast failed? Does the data is still taken and need a reload?

src/geode/basic/database.cpp
```diff
@@ -0,0 +1,345 @@
...
+        }
+
+        index_t nb_data() const
+        {
+            return storage_.size();
+        }
+
+        const uuid& register_unique_data( std::unique_ptr< Identifier >&& data )
+        {
+            const auto& registedred_data = register_data( std::move( data ) );
```
MelchiorSchuh, line 158: registered

src/geode/basic/database.cpp
MelchiorSchuh: This returns a const ref, so I'd say the dev should know it mustn't destroy Data, or else the ref points to nothing (If I understand this issue correctly).

src/geode/basic/database.cpp
BotellaA: When the data is not used for this amount of time, it is unloaded from the memory.

include/geode/basic/database.h
BotellaA: empty or exception? I would prefer exception I think

src/geode/basic/database.cpp
panquez: you mean BRep or BMesh?

src/geode/basic/database.cpp
panquez: You at least add a warning message into the docs.

tests/basic/test-database.cpp
```diff
@@ -0,0 +1,121 @@
...
+    auto& taken_foo = dynamic_cast< Foo& >( *unique_id );
+    taken_foo.value_ = 12;
+    OPENGEODE_EXCEPTION(
+        taken_foo.value_ == 12, "[Test] Wrong value after modify taken data" );
+    OPENGEODE_EXCEPTION(
+        foo.value_ == 42, "[Test] Wrong value after register data" );
+    database.register_data( std::move( unique_id ) );
+    auto foo_data2 = database.get_data( id );
+    const auto& foo2 = foo_data2.get< Foo >();
+    OPENGEODE_EXCEPTION(
```
panquez, line 106: and if you test foo here ? what happens?

src/geode/basic/database.cpp
```diff
@@ -0,0 +1,345 @@
...
+            const auto it = storage_.find( id );
+            if( it != storage_.end() )
+            {
+                if( it->second->unused() )
+                {
+                    it->second->set_data( std::move( data ) );
+                    return it->second->data();
+                }
+                delete_data( id );
+                return do_register_data( std::move( data ) );
```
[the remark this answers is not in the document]
BotellaA: if the file is already in the DB, remove it and register the new one. If there is some `Data` object, they will keep their shared_ptr to the "old" data. Only newly created `Data`will reflect the changes.

src/geode/basic/database.cpp
BotellaA: BRep, but both are not fully compatible at the moment. There is no serialize method.

tests/basic/test-database.cpp
BotellaA: foo is supposed to have the old value, Let's try!

BotellaA:
We still need python binding and testing on more complex data like BRep. Should we do it now or later?

include/geode/basic/database.h
```diff
@@ -0,0 +1,164 @@
...
+    class opengeode_basic_api Database
+    {
+        OPENGEODE_DISABLE_COPY( Database );
+        struct Storage;
+
+    public:
+        using serializer_function = std::function< void( PContext& ) >;
+
+        /*!
+         * Classe holding a const reference of a data.
```
panquez, line 55:
```suggestion
         * Class holding a const reference of data.
```

include/geode/basic/database.h
```diff
@@ -0,0 +1,164 @@
...
+         * @warning Do not destroy this Data class before the const reference
+         * obtained using its get() method is no longer in used
```
panquez, line 57:
```suggestion
         * @warning Do not destroy this Data class if the const reference
         * obtained using its get() method is still used
```
else? what's happened if it is done?

include/geode/basic/database.h
```diff
@@ -0,0 +1,164 @@
...
+            return register_unique_data( id, std::move( data ) );
+        }
+
+        void delete_data( const uuid& id );
+
+        /*!
+         * Retrieve a read only reference to the data corresponding to the given
+         * uuid.
+         */
+        Data get_data( const uuid& id ) const;
```
panquez, line 137: read only but not const ref?

include/geode/basic/database.h
BotellaA: Do we really need to document what happen if you di not respect the rule?

include/geode/basic/database.h
BotellaA: Yes, this is a wrapper, not the real data

include/geode/basic/database.h
panquez: No but can you tell me, just to know for future debug?

BotellaA: when all the Data objects are deleted, the database starts a countdown to offload the actual data from the memory. So if you destroy your Data object but are still using the referenced data, you might get segfault or weird bugs because the memory has been cleaned.

panquez (approved):

PR #540, opened 2022-09-10
fix(1D): add 1D export on basic objects and helpers

BotellaA:


include/geode/geometry/basic_objects/segment.h
```diff
@@ -82,7 +81,7 @@ namespace geode
         OwnerSegment< dimension >& operator=(
             OwnerSegment< dimension >&& other );
     };
-    ALIAS_2D_AND_3D( OwnerSegment );
+    ALIAS_1D_AND_2D_AND_3D( OwnerSegment );
```
panquez, line 84: Is direction a method of Segment1D, what does it mean?

BotellaA: Not sure.. Probably the direction on the segment like +1 or - 1

src/geode/geometry/distance.cpp
```diff
@@ -636,6 +636,15 @@ namespace geode
...
     }
 
+    template double opengeode_geometry_api point_point_distance(
+        const Point1D&, const Point1D& );
+    template std::tuple< double, Point1D > opengeode_geometry_api
+        point_segment_distance( const Point1D&, const Segment1D& );
+    template std::tuple< double, Point1D, Point1D > opengeode_geometry_api
+        segment_segment_distance( const Segment1D&, const Segment1D& );
+    template std::tuple< double, Point1D > opengeode_geometry_api
+        point_line_distance( const Point1D&, const InfiniteLine1D& );
```
panquez, line 646: i think it is always 0. useful?

src/geode/geometry/projection.cpp
```diff
@@ -72,6 +72,11 @@ namespace geode
         return point - plane.normal() * distance;
     }
 
+    template Point1D opengeode_geometry_api point_segment_projection(
+        const Point1D&, const Segment1D& );
+    template Point1D opengeode_geometry_api point_line_projection(
```
panquez, line 77: almost same remark

panquez (approved):

panquez:
Bindings todo when needed

PR #541, opened 2022-09-13
fix(BoundaingBox): missing method Python bindings

BotellaA:


MelchiorSchuh (approved):
typo in the commit name

panquez (approved):
Be careful on the typo in commit message

PR #553, opened 2022-10-04
Raystuff

panquez:


src/geode/mesh/helpers/ray_tracing.cpp
```diff
@@ -39,14 +39,123 @@
...
+    bool test_vertex_mode( const geode::SurfaceMesh3D& mesh,
+        const geode::RayTracing3D::PolygonDistance& polygon0,
+        const geode::RayTracing3D::PolygonDistance& polygon1 )
+    {
+        geode::PolygonVertex polygon_vertex0{ polygon0.polygon, geode::NO_LID };
+        if( polygon0.position == geode::Position::vertex0 )
+        {
+            polygon_vertex0.vertex_id = 0;
+        }
+        if( polygon0.position == geode::Position::vertex1 )
```
BotellaA, line 61: maybe use else if instead of multiple if to gain some perf

BotellaA (approved):

src/geode/mesh/helpers/ray_tracing.cpp
BotellaA: for edge case too

panquez (commented):
LGTM

PR #556, opened 2022-10-05
fix(PolyhedronFacetVertices): Set base number to 3 in InlinedVector.

MelchiorSchuh:


BotellaA (commented):
Why not using the alias everywhere, instead of the inlined vector.

BotellaA (approved):

BotellaA:
Go merge when you are ready

PR #561, opened 2022-10-07
fix(LineTracing): handle the negative side of the line

BotellaA:


src/geode/mesh/helpers/ray_tracing.cpp
```diff
@@ -254,13 +262,13 @@ namespace geode
...
-                            { triangle.vertices()[e2].get(),
-                                triangle.vertices()[( e2 + 1 ) % 3].get() } );
-                        auto distance = point_point_distance(
-                            segment_.vertices()[0].get(), point );
-                        if( Vector3D{ segment_.vertices()[0].get(), point }.dot(
+                        std::tie( std::ignore, point, std::ignore ) =
+                            segment_segment_distance( Segment3D{ segment_ },
+                                { triangle.vertices()[e2].get(),
+                                    triangle.vertices()[( e2 + 1 ) % 3]
+                                        .get() } );
```
panquez, line 269: maybe instead of using 2 ignore, use std::get<1>

panquez (approved):

PR #564, opened 2022-10-11
fix(SurfaceMesh): better update polygon around vertex after polygon delete

BotellaA:


src/geode/mesh/builder/surface_mesh_builder.cpp
```diff
@@ -219,6 +219,9 @@ namespace
                         break;
                     }
                 }
+            }
+            if( new_polygon_vertex.polygon_id == geode::NO_ID )
```
panquez, line 223: what is different before previous version and this one ?

panquez, line 223: What is the difference?

panquez (approved):

PR #575, opened 2022-10-19
perf(VertexIdentifier): remove useless large memory allocation

BotellaA:


src/geode/model/mixin/core/vertex_identifier.cpp
```diff
@@ -431,10 +428,9 @@ namespace geode
             const uuid& component_id )
         {
             std::vector< index_t > result;
-            result.reserve( nb_unique_vertices() );
             for( const auto uv_id : Range{ nb_unique_vertices() } )
             {
-                if( !component_mesh_vertices( uv_id, component_id ).empty() )
```
BotellaA, line 437: This is the real change:
- no allocation for the entire nb of vertices...
- component_mesh_vertices( X, X ) create a temporary vector (constructor and destructor are clearly visible with callgrind)

panquez (approved):

PR #576, opened 2022-10-19
perf(Container): prefer operator[] rather than method at

panquez:
Where do you suggest to add ASSERTs?


src/geode/geometry/distance.cpp
```diff
@@ -70,7 +70,13 @@ namespace geode
     double point_point_distance(
         const Point< dimension >& point0, const Point< dimension >& point1 )
     {
-        return Vector< dimension >{ point0, point1 }.length();
+        double diff2{ 0 };
```
BotellaA, line 73: Why does this make so much difference? It increases maintenance...

PR #577, opened 2022-10-19
fix(filename): Added a function to get the path to a file without the…

MelchiorSchuh:
… file name.


BotellaA (approved):

tests/basic/test-filename.cpp
```diff
@@ -35,18 +35,34 @@ void test()
         "[Test] Wrong result for path1 with extension" );
     OPENGEODE_EXCEPTION( geode::filename_without_extension( path1 ) == "file",
         "[Test] Wrong result for path1 without extension" );
+    OPENGEODE_EXCEPTION(
+        geode::filepath_without_extension( path1 ) == "/where/is/file",
```
BotellaA, line 39: The path starting with / will not work on windows

panquez (approved):

PR #587, opened 2022-10-31
fix(ModelHelpers): add overload for converting model meshes using bui…

BotellaA:
…lders

### DIFF
--- a/include/geode/model/helpers/convert_model_meshes.h
+++ b/include/geode/model/helpers/convert_model_meshes.h
@@ -28,21 +28,33 @@
 namespace geode
 {
     class BRep;
+    class BRepBuilder;
     class Section;
+    class SectionBuilder;
 } // namespace geode
 
 namespace geode
 {
     void opengeode_model_api convert_surface_meshes_into_triangulated_surfaces(
         BRep& brep );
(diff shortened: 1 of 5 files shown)


include/geode/model/helpers/convert_model_meshes.h
```diff
@@ -28,21 +28,33 @@
...
+    class BRepBuilder;
     class Section;
+    class SectionBuilder;
 } // namespace geode
 
 namespace geode
 {
     void opengeode_model_api convert_surface_meshes_into_triangulated_surfaces(
         BRep& brep );
+    void opengeode_model_api convert_surface_meshes_into_triangulated_surfaces(
```
BotellaA, line 40: Should we deprecate the other one? 
I think both should be kept.

include/geode/model/representation/core/brep.h
```diff
@@ -67,6 +67,8 @@ namespace geode
                                      public Identifier
     {
     public:
+        static constexpr index_t dimension{ 3 };
```
BotellaA, line 70: This is very handy for template progamming

include/geode/model/helpers/convert_model_meshes.h
panquez: I agree

panquez (approved):

PR #594, opened 2022-11-09
perf(SolidMesh): improve polyhedron_facet_edge_from_vertices

BotellaA:


panquez (commented):
Why having separate the 2 if in the first tests but not on the else one ?

BotellaA:
If the first test is true, no need to test the else, it will not work. But once your are in the else, you can split the test or not, the resulting computation flow will be the same

panquez (approved):

PR #597, opened 2022-11-10
feat(ModelHelpers): cut block mesh along internal surfaces

BotellaA:


src/geode/mesh/builder/solid_mesh_builder.cpp
```diff
@@ -280,6 +280,76 @@ namespace
...
+    template < geode::index_t dimension >
+    void update_edge_and_facet( const geode::SolidMesh< dimension >& solid,
+        geode::SolidMeshBuilder< dimension >& builder,
+        const geode::PolyhedronVertex& polyhedron_vertex,
+        geode::index_t new_vertex_id )
+    {
+        for( const auto f : geode::LRange{ solid.nb_polyhedron_facets(
+                 polyhedron_vertex.polyhedron_id ) } )
+        {
+            const geode::PolyhedronFacet id{ polyhedron_vertex.polyhedron_id,
```
panquez, line 293: facet_id instead of id ?

src/geode/mesh/builder/solid_mesh_builder.cpp
```diff
@@ -280,6 +280,76 @@ namespace
...
+                f };
+            const auto nb_facet_vertices =
+                solid.nb_polyhedron_facet_vertices( id );
+            absl::FixedArray< geode::PolyhedronVertex > facet_vertices(
+                nb_facet_vertices );
+            for( const auto v : geode::LRange{ nb_facet_vertices } )
```
panquez, line 299: Is there a function that return the result of this for loop? Seems an interesting feature.

src/geode/mesh/builder/solid_mesh_builder.cpp
```diff
@@ -280,6 +280,76 @@ namespace
...
+                for( const auto& v : facet_vertices )
+                {
+                    facet_vertices_id.emplace_back(
+                        solid.polyhedron_vertex( v ) );
+                }
+
+                const auto position = static_cast< geode::index_t >(
+                    std::distance( facet_vertices.begin(), position_it ) );
+
+                if( solid.are_facets_enabled() )
```
panquez, line 319: Maybe skip at thé begining of the function if both edges and facets are not enabled

src/geode/mesh/builder/solid_mesh_builder.cpp
```diff
@@ -339,69 +433,8 @@ namespace geode
         if( solid_mesh_.are_edges_enabled()
             || solid_mesh_.are_facets_enabled() )
         {
-            for( const auto f : LRange{ solid_mesh_.nb_polyhedron_facets(
```
panquez, line 342: Ok the skip is here? What is the best place?

panquez (approved):

src/geode/mesh/builder/solid_mesh_builder.cpp
BotellaA: I like it there

src/geode/mesh/builder/solid_mesh_builder.cpp
BotellaA: not yet. I already made the `polyhedron_facet_vertex_id` public. It will allow us to think about other APIs I am sure.

PR #599, opened 2022-11-14
fix(repair_polygon_orientations): Reset polygons_around_vertex cache …

MelchiorSchuh:
…value after reorientation.


src/geode/mesh/helpers/repair_polygon_orientations.cpp
```diff
@@ -305,6 +306,7 @@ namespace
                 {
                     builder.unset_polygon_adjacent( { p, v } );
                 }
+                builder.reset_polygons_around_vertex( bad_polygon_vertex );
```
BotellaA, line 309: This should already be handled in the unset adjacent.

BotellaA (changes requested):
Are you sure about this change?

PR #610, opened 2022-12-02
fix(Exception): handle exception in worker thread

BotellaA:


src/geode/model/mixin/core/blocks.cpp
```diff
@@ -126,11 +126,13 @@ namespace geode
                 }
             } );
         }
-        async::when_all( tasks.begin(), tasks.end() )
-            .then( [level] {
-                Logger::set_level( level );
-            } )
-            .wait();
+        auto all_tasks = async::when_all( tasks.begin(), tasks.end() );
```
panquez, line 129: Why not when_all(…).get() here?

BotellaA: If the when_all throws an exception, le Logger will not be reset to its initial logging level. It is unlikely (but possible) that the when_all throws but just to be safe.

MelchiorSchuh (approved):

PR #626, opened 2022-12-28
fix(BoundingBox): add more primitive intersections

BotellaA:


include/geode/geometry/bounding_box.h
```diff
@@ -61,6 +66,18 @@ namespace geode
 
         bool intersects( const InfiniteLine< dimension >& line ) const;
 
+        template < index_t T = dimension >
+        typename std::enable_if< T == 2, bool >::type intersects(
+            const Segment< T >& segment ) const;
+
+        template < index_t T = dimension >
+        typename std::enable_if< T == 2 || T == 3, bool >::type intersects(
```
panquez, line 74: If i have rightly understood impl, it intersects (at least in 2D) if and only if the edges (or the facets of tetra) intersect the bbox. What is the aimed behavior when the bbox is included in the triangle?


At least, these information should be clarify in documentation.

src/geode/geometry/bounding_box.cpp
```diff
@@ -128,6 +133,59 @@ namespace geode
...
+            const auto& vertices = tetra.vertices();
+            return intersects( { vertices[0].get(), vertices[1].get(),
+                       vertices[2].get() } )
+                   || intersects( { vertices[0].get(), vertices[1].get(),
+                       vertices[3].get() } )
+                   || intersects( { vertices[0].get(), vertices[2].get(),
+                       vertices[3].get() } )
+                   || intersects( { vertices[1].get(), vertices[2].get(),
+                       vertices[3].get() } );
+        }
```
panquez, line 153: || bbox.center() is in the tetrahedron

include/geode/geometry/bounding_box.h
BotellaA: The aimed behavior is intersection if the element is crossing, is inside, or is containing the bbox. (I am adding this in the doc)

The implementation does not really detect intersections, it uses the separating axis method which detects overlapping intervals of the projected edge (or facet) and the bbox on several axes. It matches the expected behavior. Tests include this special case.

src/geode/geometry/bounding_box.cpp
BotellaA: No need. The intersections tests are enough

panquez (approved):

PR #628, opened 2023-01-05
fix(BoundingBox): when checking intersection with 3D triangles check …

panquez:
…first their vertices

### DIFF
--- a/src/geode/geometry/bounding_box.cpp
+++ b/src/geode/geometry/bounding_box.cpp
@@ -148,6 +148,13 @@ namespace geode
                 return true;
             }
             const auto& vertices = tetra.vertices();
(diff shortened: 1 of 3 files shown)


MelchiorSchuh (approved):

src/geode/geometry/bounding_box.cpp
```diff
@@ -326,6 +326,13 @@ namespace geode
         const Triangle< 3 >& triangle ) const
     {
         const auto& vertices = triangle.vertices();
+        for( const auto v : LRange{ 3 } )
```
BotellaA, line 329: Should we do the same for Tetra and Triangle2d ?

BotellaA (approved):

PR #638, opened 2023-01-13
fix(ModelHelpers): missing model meshes AABBTrees

BotellaA:
We already add some model helpers to create the model components AABBTree but without the mesh trees. 

Do you think this new API is useful or redundant? I think it simplifies the use and it is the goal of the helpers.

### DIFF
--- a/include/geode/model/helpers/aabb_model_helpers.h
+++ b/include/geode/model/helpers/aabb_model_helpers.h
@@ -24,6 +24,9 @@
 #pragma once
 
 #include <absl/container/fixed_array.h>
+#include <absl/container/flat_hash_map.h>
+
+#include <geode/basic/uuid.h>
 
 #include <geode/model/common.h>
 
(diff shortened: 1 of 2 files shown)


panquez (approved):

include/geode/model/helpers/aabb_model_helpers.h
```diff
@@ -52,4 +54,36 @@ namespace geode
...
+
+        AABBTree< dimension > model_tree_;
+        absl::FixedArray< AABBTree< dimension > > meshes_trees_;
+        absl::FixedArray< uuid > uuids_;
+        absl::flat_hash_map< uuid, index_t > tree_ids_;
+    };
+    ALIAS_2D_AND_3D( ModelMeshesAABBTree );
+
+    ModelMeshesAABBTree3D opengeode_model_api
+        create_lines_and_meshes_aabb_trees( const BRep& model );
```
MelchiorSchuh, line 75: This is more create_lines_aabb_trees, since you don't create aabb trees for other meshes (idem for the others)

PR #639, opened 2023-01-13
fix(segment_triangle_intersection_detection): Added management of cop…

MelchiorSchuh:
…lanar case.


BotellaA:
Are there enough tests for this coplanar case?

PR #642, opened 2023-01-17
fix(intersection-detection): return outside in case of a segment touching the triangle on one vertex in segment_triangle_intersection.

MelchiorSchuh:


src/geode/geometry/intersection_detection.cpp
```diff
@@ -91,23 +99,65 @@ namespace
...
+            if( ( seg_edge_inter.first == geode::Position::vertex0
+                    || seg_edge_inter.first == geode::Position::vertex1 )
+                && ( seg_edge_inter.second == geode::Position::vertex0
+                     || seg_edge_inter.second == geode::Position::vertex1 ) )
             {
-                return { geode::Position::parallel, geode::Position::parallel };
+                result.first = seg_edge_inter.first;
+                if( seg_edge_inter.second == geode::Position::vertex0 )
+                {
+                    if( edge_v0 == 0 )
```
BotellaA, line 115: You could create a static const array of Position to avoid the if/elseif/else block

BotellaA (approved):

src/geode/geometry/intersection_detection.cpp
```diff
@@ -34,18 +34,27 @@
...
             }
         }
-        return axis_to_remove;
+        if( largest_axis == 0 )
+        {
+            return { 1, 2 };
+        }
+        if( largest_axis == 1 )
+        {
+            return { 0, 2 };
```
panquez, line 55: 2, 0 ?

panquez: to keep the order

panquez (approved):

PR #646, opened 2023-01-26
fix(create_model_aabb_trees): Removed race condition introducing errors when creating model component trees.

MelchiorSchuh:


src/geode/model/helpers/aabb_model_helpers.cpp
```diff
@@ -82,10 +82,10 @@ namespace
             tasks[id] = async::spawn( [id, &result, &boxes, &element] {
                 result.mesh_trees_[id] =
                     geode::create_aabb_tree( element.mesh() );
-                result.uuids_[id] = element.id();
-                result.mesh_tree_ids_.emplace( element.id(), id );
                 boxes[id] = result.mesh_trees_[id].bounding_box();
             } );
+            result.uuids_[id] = element.id();
```
BotellaA, line 87: I think this one is an array so it is fine to leave it in the async task.

BotellaA (approved):

PR #648, opened 2023-01-31
add thickness parameter to convert 2D object to 3D

francoisbonneau:


src/geode/model/helpers/convert_brep_section.cpp
```diff
@@ -121,23 +121,59 @@ namespace geode
...
         }
         copy_unique_vertices( section, builder, mappings );
         return std::make_tuple( std::move( brep ), std::move( mappings ) );
     }
+
+    std::tuple< BRep, ModelCopyMapping > extrude_section_to_brep(
+        const Section& section,
+        index_t axis_to_add,
+        double min_axis_coordinate,
+        double thickness )
```
francoisbonneau, line 151: Should we use min and max values? or min value and extrusion thickness.

francoisbonneau:
Should we do so many changes in the API?

src/geode/model/helpers/convert_brep_section.cpp
panquez: Maybo a top and bottom depth is a good idea

francoisbonneau: bottom and top are maybe too specific.

PR #651, opened 2023-02-03
fix(Cylinder): add new geometrical basic objects: the cylinder

panquez:


include/geode/geometry/basic_objects/cylinder.h
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+#pragma once
+
+#include <geode/geometry/basic_objects/segment.h>
+#include <geode/geometry/common.h>
+
+namespace geode
+{
+    using Basis3D = std::array< Vector3D, 3 >;
```
BotellaA, line 31: Could this alias be defined inside the Cylinder class?

BotellaA (approved):

include/geode/geometry/basic_objects/cylinder.h
MelchiorSchuh: Basis or Base ?

include/geode/geometry/basic_objects/cylinder.h
```diff
@@ -0,0 +1,51 @@
...
+        Cylinder& operator=( Cylinder&& other );
+
+        const Segment3D& axis() const;
+        const Basis3D& basis() const;
+        double radius() const;
+
+    private:
+        Segment3D axis_;
+        double radius_{ 0 };
+        Basis3D basis_;
```
MelchiorSchuh, line 49: Should this really be a class object, or should it be computed only when needed (and eventually stored in a CacheData) ?

include/geode/geometry/basic_objects/cylinder.h
panquez: I've found bsis on wikipedia page about Cylinder definition

include/geode/geometry/intersection.h
```diff
@@ -39,7 +39,9 @@ namespace geode
     ALIAS_2D_AND_3D( Point );
     ALIAS_2D_AND_3D( Segment );
     ALIAS_3D( Triangle );
+    ALIAS_3D( Triangle );
```
MelchiorSchuh, line 42: I think you forgot a copy

include/geode/geometry/basic_objects/cylinder.h
panquez: Good question, any thought @BotellaA?

src/geode/geometry/intersection.cpp
```diff
@@ -24,15 +24,32 @@
...
 #include <geode/geometry/basic_objects/sphere.h>
 #include <geode/geometry/basic_objects/triangle.h>
 #include <geode/geometry/distance.h>
 #include <geode/geometry/perpendicular.h>
+#include <geode/geometry/projection.h>
 #include <geode/geometry/vector.h>
 
+namespace
+{
+    struct CyclinderLineResult
```
MelchiorSchuh, line 40:
```suggestion
    struct CylinderLineResult
```

MelchiorSchuh: Here and everywhere it is used

src/geode/geometry/intersection.cpp
```diff
@@ -494,6 +491,327 @@ namespace geode
         return line_intersection_result.type;
     }
 
+    IntersectionResult< absl::InlinedVector< Point3D, 2 > >
```
MelchiorSchuh, line 494: Maybe cut this into several sub-functions depicting the different cases, with variables with more meaningful names

panquez: I agree but it would be preferable to conserve the initial code

include/geode/geometry/intersection.h
```diff
@@ -172,4 +174,28 @@ namespace geode
...
+    IntersectionResult< absl::InlinedVector< Point3D, 2 > >
+        opengeode_geometry_api line_cylinder_intersection(
+            const InfiniteLine3D& line, const Cylinder& cylinder );
+
+    /*!
+     * Compute the intersection between a segment and a cylinder
+     * @return an optional of the intersection points.
+     */
+    IntersectionResult< absl::InlinedVector< Point3D, 2 > >
+        opengeode_geometry_api segment_cylinder_intersection(
```
BotellaA, line 199: Same as L183

include/geode/geometry/basic_objects/cylinder.h
BotellaA: Is it really Cylinder related or Segment/Cylinder intersection related? maybe move it in intersection.cpp?

PR #665, opened 2023-02-28
fix(CMake): provide target with less tests

BotellaA:
To use it, I will change the actions to:
```
cd build
make third_party
cd opengeode
make light
make test
```

What tests are we enabling?

### DIFF
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -123,6 +123,9 @@ if(DOXYGEN_FOUND AND EXISTS ${PROJECT_SOURCE_DIR}/cmake/Doxyfile.in)
         VERBATIM )
 endif()
 
+add_custom_target(third_party)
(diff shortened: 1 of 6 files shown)


cmake/utils.cmake
```diff
@@ -123,6 +123,9 @@ if(DOXYGEN_FOUND AND EXISTS ${PROJECT_SOURCE_DIR}/cmake/Doxyfile.in)
         VERBATIM )
 endif()
 
+add_custom_target(third_party)
+add_custom_target(light)
```
francoisbonneau, line 127: est ce que les nom de targets  "light_test" ou "set_light_test" ne serait pas plus explict?

francoisbonneau (approved):

MelchiorSchuh (approved):

panquez (approved):

PR #667, opened 2023-03-01
feat(ModelConcatener): Added a class to concatenate a brep to another.

MelchiorSchuh:


include/geode/model/helpers/model_concatener.h
```diff
@@ -0,0 +1,54 @@
...
+namespace geode
+{
+    class BRep;
+    class Section;
+} // namespace geode
+
+namespace geode
+{
+    template < typename Model >
+    class opengeode_model_api ModelConcatener
```
BotellaA, line 39:
```suggestion
    class ModelConcatener
```
This is a templated class only exported in the cpp file

src/geode/model/helpers/model_concatener.cpp
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+    private:
+        void copy_relationships(
+            const Model& other_model, const ModelCopyMapping& mapping );
+
+        void copy_line_relationships(
+            const Model& other_model, const ModelCopyMapping& mapping )
+        {
+            for( const auto& line : other_model.lines() )
```
BotellaA, line 156: You could get the line_mapping and corner_mapping once a the beginning of the method to avoid having the mapping.at() each time

This could be apply for every methods

BotellaA (approved):
Should we add python binding for this?

panquez (approved):

PR #669, opened 2023-03-02
2D to 3D converter

francoisbonneau:
BREAKING CHANGE:
feat(2d to 3d converters): add a parameter.
Now, you can choose the value of the coordinate to add.


BotellaA:
Commit message should be:
```
feat(Helpers): add a parameter to convert object from 2D to 3D

BREAKING CHANGE: Now, you can choose the value of the coordinate to add.
```
The order of your message is not correct

BotellaA (approved):

MelchiorSchuh (approved):

PR #683, opened 2023-03-15
fix(BoundingBox): handle degenerated segment intersection

BotellaA:


src/geode/geometry/nn_search.cpp
```diff
@@ -185,6 +186,9 @@ namespace geode
     typename NNSearch< dimension >::ColocatedInfo
         NNSearch< dimension >::colocated_index_mapping( double epsilon ) const
     {
+        OPENGEODE_EXCEPTION( epsilon >= global_epsilon,
+            "[NNSearch::colocated_index_mapping] Given epsilon too small, "
+            "should be bigger than global_epsilon" );
```
panquez, line 191:
```suggestion
            "should be bigger than global_epsilon (i.e. ", global_epsilon, ")" );
```

panquez (approved):

PR #684, opened 2023-03-16
fix(SurfaceMerger): fix adjacency issues when merging surfaces with d…

panquez:
…uplicated polygons

### DIFF
--- a/tests/mesh/test-merge-surface.cpp
+++ b/tests/mesh/test-merge-surface.cpp
@@ -36,7 +36,7 @@ void test()
 {
     geode::OpenGeodeMesh::initialize();
     std::vector< geode::Point2D > points{ { { 0, 0 } }, { { 0, 1 } },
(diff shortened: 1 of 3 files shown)


BotellaA (approved):

tests/mesh/test-merge-surface.cpp
```diff
@@ -66,25 +72,41 @@ void test()
         merged->nb_vertices() == 6, "[Test] Wrong number of vertices" );
     OPENGEODE_EXCEPTION(
         merged->nb_polygons() == 4, "[Test] Wrong number of polygons" );
+    for( const auto p : geode::Range{ merged->nb_polygons() } )
+    {
+        for( const auto e : geode::LRange{ 3 } )
+        {
+            merged->polygon_adjacent_edge( { p, e } );
+        }
+    }
```
MelchiorSchuh, line 81: I'm not sure this double loop does anything ... is there an exception missing ?

panquez: There are internal tests but i do not know if some compiler will remove this instruction.

PR #721, opened 2023-04-26
CMake MultiConfig

BotellaA:


cmake/utils.cmake
```diff
@@ -467,22 +468,18 @@ print(name + version + '-' + abi + '-' + platform)"
...
-            COMMAND ${CMAKE_COMMAND} -E copy_directory "${wheel_build_directory}/${binary_folder}" "${wheel_output_directory}/${binary_folder}"
-            COMMAND ${CMAKE_COMMAND} -E copy_directory "${wheel_build_directory}/share" "${wheel_output_directory}/share"
-            COMMAND ${CMAKE_COMMAND} -E remove "${wheel_output_directory}/${binary_folder}/*.py"
-            COMMAND ${PYTHON_EXECUTABLE} setup.py bdist_wheel
-            WORKING_DIRECTORY ${PROJECT_BINARY_DIR}/wheel
-        )
+        set(wheel_config_folder "$<$<BOOL:${isMultiConfig}>:$<CONFIG>>")
     endif()
+    add_custom_target(wheel
+        COMMAND ${CMAKE_COMMAND} -E copy_directory "${wheel_build_directory}/${binary_folder}/${wheel_config_folder}" "${wheel_output_directory}/${binary_folder}"
```
MelchiorSchuh, line 477: Are you sure you won't have a // instead of / if wheel_config_folder="" ? Will it be a problem if you do ?

MelchiorSchuh (approved):

cmake/utils.cmake
BotellaA: // is not an issue

PR #724, opened 2023-04-28
fix(Range): simplify for MSVC compilation

BotellaA:


cmake/utils.cmake
```diff
@@ -21,6 +21,7 @@ include(InstallRequiredSystemLibraries)
 
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
+set(CMAKE_COLOR_DIAGNOSTICS ON)
```
BotellaA, line 24: @MelchiorSchuh this is to activate color in the build log

include/geode/basic/range.h
```diff
@@ -93,25 +93,16 @@ namespace geode
      * Begin and end methods for range-based iteration on custom range.
      * See derived classes for usage.
      */
-    template < typename Type >
```
panquez, line 96: to be kept but deprecated

panquez (approved):

MelchiorSchuh (approved):

PR #727, opened 2023-05-05
fix(Intersection): missing Circle intersections

BotellaA:


src/geode/geometry/intersection.cpp
```diff
@@ -838,14 +839,172 @@ namespace geode
...
+        // where det = 1 - d^2.
+        const auto dot = plane0.normal().dot( plane1.normal() );
+        const auto constant0 = plane0.plane_constant();
+        const auto constant1 = plane1.plane_constant();
+        if( std::fabs( dot ) >= 1. )
+        {
+            // The planes are parallel.  Check if they are coplanar.
+            const auto constant_diff =
+                dot >= 0. ? constant0 - constant1 : constant0 + constant1;
+            if( std::fabs( constant_diff ) == 0. )
```
panquez, line 981:
```suggestion
            if( constant_diff == 0. )
```
could it fail if it returns -0?

src/geode/geometry/intersection.cpp
```diff
@@ -838,14 +839,172 @@ namespace geode
...
+                correctness.second.second.emplace_back(
+                    plane_intersection.correctness->second.second[i] );
+            }
+        }
+        if( result.empty() )
+        {
+            return { IntersectionType::NONE };
+        }
+        correctness.first.first = true;
+        correctness.second.first = true;
```
panquez, line 888: i do not understand?

BotellaA: L887: all points are in the thick triangle because of test L874
L888: all points are in the thick circle or test on L861 will have already failed

src/geode/geometry/intersection.cpp
BotellaA: I do not know, took this code from geometricTools

panquez (approved):

PR #731, opened 2023-05-11
fix(Circle): missing bounding box

BotellaA:


src/geode/geometry/basic_objects/circle.cpp
```diff
@@ -67,6 +69,33 @@ namespace geode
...
+        const auto angle = [this]( Vector3D direction ) {
+            double dP = plane_.normal().dot( direction );
+            if( dP <= -1.0 )
+            {
+                return M_PI;
+            }
+            if( dP >= 1.0 )
+            {
+                return 0.;
+            }
```
panquez, line 84: if direction and normal are normalized, is this possible / needed?

src/geode/geometry/basic_objects/circle.cpp
```diff
@@ -67,6 +69,33 @@ namespace geode
...
+
+            return std::acos( dP );
+        };
+        const auto angle_x = angle( { { 1, 0, 0 } } );
+        const auto angle_y = angle( { { 0, 1, 0 } } );
+        const auto angle_z = angle( { { 0, 0, 1 } } );
+        const auto translation = Vector3D{
+            { std::sin( angle_x ), std::sin( angle_y ), std::sin( angle_z ) }
```
panquez, line 92: sin(arccosx) = sqrt(1-x^2)

panquez (approved):
missing python bindings and tests

src/geode/geometry/basic_objects/circle.cpp
BotellaA: it only works if they are normalized but it is the case here

PR #733, opened 2023-05-12
fix(ComponentMeshElement): Added methods to get a polyhedron from the unique vertices of its vertices in a model.

MelchiorSchuh:


src/geode/model/mixin/core/vertex_identifier.cpp
```diff
@@ -87,6 +93,10 @@ namespace geode
         const std::vector< ComponentMeshVertex >& component_mesh_vertices(
             index_t unique_vertex_id ) const
         {
+            OPENGEODE_EXCEPTION( unique_vertex_id < nb_unique_vertices(),
```
BotellaA, line 96: Maybe an assertion?

PR #739, opened 2023-05-16
fix(Model): getter to model meshes crss

BotellaA:


JulienChampagnol (approved):

MelchiorSchuh (approved):

src/geode/model/helpers/model_coordinate_reference_system.cpp
```diff
@@ -0,0 +1,99 @@
...
+            crss.emplace(
+                crs_manager.active_coordinate_reference_system().type_name(),
+                crs_manager.active_coordinate_reference_system_name() );
+        }
+    }
+
+    absl::FixedArray< std::pair< geode::CRSType, std::string > > get_result(
+        absl::flat_hash_set< std::pair< geode::CRSType, std::string > >& crss )
+    {
+        absl::FixedArray< std::pair< geode::CRSType, std::string > > result(
```
panquez, line 66: Why do you prefer to return a fixed array instead of a set?

panquez (approved):

src/geode/model/helpers/model_coordinate_reference_system.cpp
BotellaA: To simplify python use

PR #740, opened 2023-05-18
fix(Build): reduce binary size

BotellaA:


include/geode/basic/attribute_manager.h
```diff
@@ -246,15 +246,6 @@ namespace geode
             context.registerSingleBaseBranch< Serializer, AttributeBase,
                 SparseAttribute< Type > >(
                 absl::StrCat( "SparseAttribute", name ).c_str() );
-            context.registerSingleBaseBranch< Serializer,
```
BotellaA, line 249: The idea is: we should not use ReadOnlyAttribute inside bitsery. 

For a savable data model, I do not think having an attribute as class member without knowing its storage impl is a good pratice.

panquez (approved):

PR #741, opened 2023-05-19
perf(Model): improve performances of component mesh vertex pairs computation

panquez:


src/geode/model/helpers/component_mesh_vertices.cpp
```diff
@@ -49,9 +49,18 @@ namespace
                     OPENGEODE_ASSERT( cmv0.vertex != cmv1.vertex,
                         "[component_mesh_vertex_pairs] Vertices should be "
                         "different" );
-                    result[cmv0.component_id].emplace_back(
-                        std::array< geode::index_t, 2 >{
-                            cmv0.vertex, cmv1.vertex } );
+                    if( result.contains( cmv0.component_id ) )
```
BotellaA, line 52: Using find and adding directly to the iterator would be faster than contains + at

BotellaA (approved):

src/geode/model/helpers/component_mesh_vertices.cpp
```diff
@@ -49,9 +49,19 @@ namespace
                     OPENGEODE_ASSERT( cmv0.vertex != cmv1.vertex,
                         "[component_mesh_vertex_pairs] Vertices should be "
                         "different" );
-                    result[cmv0.component_id].emplace_back(
-                        std::array< geode::index_t, 2 >{
-                            cmv0.vertex, cmv1.vertex } );
+                    auto itr = result.find( cmv0.component_id );
```
BotellaA, line 52: It looks weird but it can be const too 😜

PR #744, opened 2023-05-23
fix(Intersection Detection): fix coplanar and generated cases for seg…

panquez:
…ment_triangle_intersection_detection

### DIFF
--- a/src/geode/geometry/intersection_detection.cpp
+++ b/src/geode/geometry/intersection_detection.cpp
@@ -26,6 +26,7 @@
 #include <geode/geometry/basic_objects/infinite_line.h>
 #include <geode/geometry/basic_objects/segment.h>
 #include <geode/geometry/basic_objects/triangle.h>
+#include <geode/geometry/bounding_box.h>
 #include <geode/geometry/intersection.h>
 #include <geode/geometry/point.h>
 #include <geode/geometry/position.h>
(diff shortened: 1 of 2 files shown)


src/geode/geometry/intersection_detection.cpp
```diff
@@ -40,15 +41,34 @@ namespace
     };
 
     std::array< geode::local_index_t, 2 > best_projection_axis(
-        const geode::Vector3D& normal )
+        const geode::Triangle3D& triangle, const geode::Segment3D& segment )
     {
+        const auto normal_result = triangle.normal();
         geode::local_index_t largest_axis{ 2 };
-        for( const auto other_axis : geode::LRange{ 2 } )
+        if( normal_result )
```
BotellaA, line 48:
```suggestion
        if( const auto normal_result = triangle.normal() )
```

src/geode/geometry/intersection_detection.cpp
```diff
@@ -40,15 +41,34 @@ namespace
     };
 
     std::array< geode::local_index_t, 2 > best_projection_axis(
-        const geode::Vector3D& normal )
+        const geode::Triangle3D& triangle, const geode::Segment3D& segment )
     {
+        const auto normal_result = triangle.normal();
```
BotellaA, line 46:
```suggestion
```

BotellaA (approved):

PR #758, opened 2023-06-06
fix(IO): strip filename white spaces

BotellaA:


panquez (approved):
Could you add a test with filename loading with whitespace?

PR #759, opened 2023-06-06
feat(Bitsery): handle dynamic bitsery registration

BotellaA:


include/geode/model/mixin/core/detail/components_storage.h
```diff
@@ -153,27 +153,6 @@ namespace geode
                 return mapping;
             }
 
-        protected:
```
BotellaA, line 156: This is kind of a breaking change but it will only affect deep code and it is in an advanced header!

panquez (approved):

PR #762, opened 2023-06-07
Fix/add_generic_build_model_boundaries_method

MelchiorSchuh:


panquez (approved):

include/geode/model/helpers/detail/build_model_boundaries.h
```diff
@@ -0,0 +1,36 @@
...
+#pragma once
+
+#include <geode/model/common.h>
+
+namespace geode
+{
+    namespace detail
+    {
+        template < class Model >
+        void build_model_boundaries(
```
BotellaA, line 33: Why having complicated template if you specialize them ? why not two overloads, one with BRep and one with Section ?

src/geode/model/helpers/detail/build_model_boundaries.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include <geode/model/representation/builder/brep_builder.h>
+#include <geode/model/representation/builder/section_builder.h>
+#include <geode/model/representation/core/brep.h>
+#include <geode/model/representation/core/section.h>
+
+namespace
+{
+    template < class Model >
+    bool is_part_of_a_model_boundary(
+        const Model& model, const geode::uuid component_id )
```
BotellaA, line 38:
```suggestion
        const Model& model, const geode::uuid& component_id )
```

BotellaA (approved):

PR #768, opened 2023-06-13
fix(SurfaceMerger): separate merged crossing surfaces

panquez:


src/geode/mesh/helpers/detail/surface_merger.cpp
```diff
@@ -239,6 +273,18 @@ namespace geode
                         }
                     }
                 }
+                for( const auto& edge :
+                    ::edge_to_polygons_around( this->mesh() ) )
+                {
+                    if( edge.second.size() == 2 )
```
BotellaA, line 279:
```suggestion
                    if( edge.second.size() < 3 )
```

src/geode/mesh/helpers/detail/surface_merger.cpp
```diff
@@ -33,6 +33,40 @@
...
+
+    template < geode::index_t dimension >
+    absl::flat_hash_map< Edge, std::vector< geode::PolygonEdge > >
+        edge_to_polygons_around( const geode::SurfaceMesh< dimension >& mesh )
+    {
+        absl::flat_hash_map< Edge, std::vector< geode::PolygonEdge > >
+            polygons_around_edges;
+        for( const auto polygon_id : geode::Range{ mesh.nb_polygons() } )
+        {
+            for( const auto polygon_edge_id :
```
BotellaA, line 48: It will be faster to get the polygon_vertices() and then create the Edge

BotellaA (approved):

PR #769, opened 2023-06-13
fix(ConvertMesh): Added some missing helpers to convert some meshes and to convert only one component mesh of a model.

MelchiorSchuh:


src/geode/mesh/helpers/convert_solid_mesh.cpp
```diff
@@ -43,6 +45,27 @@ namespace
         }
         return true;
     }
+
+    bool all_polyhedra_are_hybrid( const geode::SolidMesh3D& solid )
+    {
+        for( const auto p : geode::Range{ solid.nb_polyhedra() } )
+        {
+            const auto nb_p_vertices = solid.nb_polyhedron_vertices( p );
+            if( nb_p_vertices < 4 || nb_p_vertices > 8 || nb_p_vertices == 7 )
```
BotellaA, line 54: It is not only 8 vertices, but 8 vertices + 6 facets of 4 vertices for an hexahedron. Same comment for the prism.

panquez (approved):

src/geode/mesh/helpers/convert_solid_mesh.cpp
panquez: it is check after i suppose with the nb_polyhedron_facet_vertices()

panquez: but yes it is true Arnaud, since it could be a polyhedron with 12 facets of 3 vertices and it passes the both tests

BotellaA (approved):

PR #784, opened 2023-07-05
perf(Distance): improve point_triangle_signed_distance3D

BotellaA:


src/geode/geometry/distance.cpp
```diff
@@ -939,19 +948,19 @@ namespace geode
...
+            const auto& nearest_point = std::get< 1 >( output );
+            const Vector3D proj2point{ nearest_point, point };
+            if( proj2point.dot( pivot_and_normal->second ) <= 0 )
+            {
+                return output;
+            }
+            return std::make_tuple( -std::get< 0 >( output ), nearest_point );
         }
-        return std::make_tuple( distance, nearest_point );
+        return no_pivot_point_triangle_distance( point, triangle );
```
panquez, line 963: I do not understand: the - side is towards the normal side and the + side opposite? Strange definition or I've missed something?

panquez (approved):

src/geode/geometry/distance.cpp
BotellaA: indeed but I did not change this part

MelchiorSchuh (approved):

PR #796, opened 2023-07-18
fix(ModelComponents): Added functions to create Components with a giv…

MelchiorSchuh:
…en uuid.

### DIFF
--- a/include/geode/model/representation/builder/detail/copy.h
+++ b/include/geode/model/representation/builder/detail/copy.h
@@ -66,6 +66,32 @@ namespace geode
         }
 
         template < typename ModelFrom, typename BuilderTo >
+        void copy_corner_components(
+            const ModelFrom& from, BuilderTo& builder_to, Mapping& mapping )
+        {
+            for( const auto& corner : from.corners() )
+            {
+                if( mapping.has_mapping_input( corner.id() ) )
+                {
+                    const auto& id = mapping.in2out( corner.id() );
+                    ModelFrom::dim == BuilderTo::dim
+                        ? builder_to.add_corner( id, corner.mesh().impl_name() )
+                        : builder_to.add_corner( id );
+                    builder_to.set_corner_name( id, corner.name() );
+                }
+                else
+                {
+                    const auto& id =
+                        ModelFrom::dim == BuilderTo::dim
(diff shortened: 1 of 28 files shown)


include/geode/model/representation/builder/detail/copy.h
```diff
@@ -65,6 +65,37 @@ namespace geode
...
+        template < typename ModelFrom, typename BuilderTo >
+        void copy_corner_components(
+            const ModelFrom& from, BuilderTo& builder_to, Mapping& mapping )
+        {
+            for( const auto& corner : from.corners() )
+            {
+                if( mapping.has_mapping_input( corner.id() ) )
+                {
+                    const auto& id = mapping.in2out( corner.id() );
+                    if( ModelFrom::dim == BuilderTo::dim )
```
BotellaA, line 77: Using different syntax here and in the else is confusing

include/geode/model/representation/builder/detail/copy.h
```diff
@@ -117,6 +211,37 @@ namespace geode
...
+                        builder_to.add_block( id );
+                    }
+                    builder_to.set_block_name( id, block.name() );
+                }
+                else
+                {
+                    const auto& id =
+                        ModelFrom::dim == BuilderTo::dim
+                            ? builder_to.add_block( block.mesh().impl_name() )
+                            : builder_to.add_block();
```
panquez, line 238: is impl_name dependant of the dimension?

BotellaA (approved):

include/geode/model/representation/core/mapping.h
```diff
@@ -71,6 +71,11 @@ namespace geode
             mappings.erase( type );
         }
 
+        MappingType& operator[]( const ComponentType& type )
+        {
+            return mappings[type];
+        }
+
```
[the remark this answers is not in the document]
MelchiorSchuh: But [] will create the element if it does not exist in the map, which is what I want

include/geode/model/representation/builder/detail/copy.h
MelchiorSchuh: Yes, impl_name will return the MeshImpl, which is dependant on the dimension (ex: OpenGeodePointSet2D)

PR #802, opened 2023-08-01
fix(Attribute): use default value in VariableAttribute copy method to…

panquez:
… allow its use with no default constructor type


BotellaA (approved):
Do the same for VariableAttribute<bool>

MelchiorSchuh (approved):

PR #803, opened 2023-08-03
Ixf/parrpttripos2d

panquez:


tests/geometry/test-position.cpp
```diff
@@ -218,6 +218,32 @@ void test_point_segment_position()
...
+void test_point_triangle_degenerated_position_2d()
+{
+    const geode::Point2D a{ { 1.0, 1.0 } };
+    const geode::Point2D b{ { 0.5, 0.5 } };
+    const geode::Point2D c{ { 1.0, 1.0 } };
+    const geode::Triangle2D triangle2D{ a, b, c };
+
+    const geode::Point2D q1{ { 1.5, 1.5 } };
+    const auto position_q1 = geode::point_triangle_position( q1, triangle2D );
+    OPENGEODE_EXCEPTION( position_q1 == geode::Position::parallel,
```
MelchiorSchuh, line 230: I would have made sure it returns geode::Position::outside, if possible

panquez: Yes I intended to do so, but it is impossible to make the difference between Position::inside et Position::outside. So I prefer to return something different to inform client that we cannot be sure.

Maybe I can try to see if I can return either Position::outside or Position::edgex

MelchiorSchuh (approved):

PR #810, opened 2023-08-09
fix(LightRegularGrid): Added missing conversion and function to get cell local vertex

MelchiorSchuh:


src/geode/mesh/core/light_regular_grid.cpp
```diff
@@ -55,6 +55,18 @@ namespace geode
             vertex_attribute_manager_.resize( nb_vertices );
         }
 
+    private:
+        friend class bitsery::Access;
+        template < typename Archive >
+        void serialize( Archive& archive )
```
BotellaA, line 61: Ok, do you also need a save/load function? Could be done later

BotellaA (approved):

PR #819, opened 2023-08-14
fix(ProgressLogger): handle multiple ProgressLoggers

BotellaA:


panquez (approved):

src/geode/basic/console_progress_logger_client.cpp
```diff
@@ -23,68 +23,88 @@
...
-            Logger::info( message_, " failed in ", timer_.duration() );
+            const auto& info = info_.at( progress_logger_id );
+            Logger::info( info.message, " failed in ", info.timer.duration() );
+            info_.erase( progress_logger_id );
         }
 
     private:
-        Timer timer_;
-        std::string message_;
+        absl::flat_hash_map< uuid, Info > info_;
```
panquez, line 80: Reading and writing on this map should be protected by mutex, should not they?

PR #826, opened 2023-08-24
fix(Rasterize): fix degenerated cases

panquez:


src/geode/mesh/helpers/rasterize.cpp
```diff
@@ -705,41 +711,88 @@ namespace
...
+        geode::index_t compute_i_coordinates( const geode::Point2D& point,
+            const geode::Vector3D& normal,
+            geode::Position position ) const
+        {
+            std::array< geode::local_index_t, 3 > order{ 0, 1, 2 };
+            if( !counter_clockwise_ )
+            {
+                order = { 0, 2, 1 };
+            }
+            if( position == geode::Position::vertex0 )
```
BotellaA, line 739: const auto floor_fabs = [](double value){return std::floor(std::fabs(value));};
Just a proposal to avoid writing the call of floor and fabs each time.
Simple lambda to replace it. As you want!

BotellaA (approved):

PR #841, opened 2023-09-22
feat(Exception): add new exception

BotellaA:
What do you think about returning geometric info at the location of the exception?

### DIFF
--- a/include/geode/geometry/point.h
+++ b/include/geode/geometry/point.h
@@ -236,7 +236,44 @@ namespace geode
             return dimension;
         }
     };
+
+    template < index_t dimension >
+    class OpenGeodePointException : public OpenGeodeException
+    {
+    public:
+        template < typename... Args >
+        explicit OpenGeodePointException(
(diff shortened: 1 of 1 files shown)


MelchiorSchuh (approved):
Not sure

include/geode/geometry/point.h
```diff
@@ -236,8 +236,30 @@ namespace geode
...
+
+    template < index_t dimension >
+    class OpenGeodePointException : public OpenGeodeException
+    {
+    public:
+        template < typename... Args >
+        explicit OpenGeodePointException(
+            Point< dimension > point_in, const Args &...message )
+            : OpenGeodeException{ absl::StrCat( message... ) },
+              point{ std::move( point_in ) }
```
panquez, line 248: Does the point coordinates be shown in the message by default ?

BotellaA: Yes, see the last version

panquez (approved):

PR #851, opened 2023-10-05
fix(Position): add mapping from Position to local_index_t

BotellaA:


include/geode/geometry/information.h
```diff
@@ -64,4 +66,10 @@ namespace geode
         // Parallel or coplanar configuration
         parallel
     };
+
+    namespace detail
+    {
+        local_index_t opengeode_geometry_api poistion_to_index(
```
panquez, line 72:
```suggestion
        local_index_t opengeode_geometry_api position_to_index(
```

src/geode/geometry/information.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include <geode/geometry/information.h>
+
+#include <absl/container/flat_hash_map.h>
+
+namespace geode
+{
+    namespace detail
+    {
+        local_index_t poistion_to_index( Position position )
```
panquez, line 32:
```suggestion
        local_index_t position_to_index( Position position )
```

panquez (approved):

PR #853, opened 2023-10-06
fix(Mapping): Added move constructors to various classes, including M…

MelchiorSchuh:
…appings.


include/geode/model/representation/core/mapping.h
```diff
@@ -42,6 +42,10 @@ namespace geode
     public:
         using Mapping = MappingType;
 
+        ModelMapping() = default;
+        ModelMapping( ModelMapping&& other ) = default;
+        ModelMapping& operator=( ModelMapping&& other ) = default;
+
```
BotellaA, line 48: Still needed? If so, what about the copy constructor? delete or default?

BotellaA (approved):

include/geode/model/representation/core/mapping.h
MelchiorSchuh: Not needed but I think they clarify the use, and they might clarify the compilation issues on the next shady uses.
I think we can delete the copy constructors, since the member is not copy-constructible

panquez (approved):

src/geode/model/helpers/convert_brep_section.cpp
```diff
@@ -114,9 +115,8 @@ namespace
             const geode::ComponentType& brep_component_type,
             geode::uuid section_component )
         {
-            return slice0_brep_mapping_.at( brep_component_type )
-                .in2out( section_slice0_mapping_.at( brep_component_type )
-                             .in2out( section_component ) );
+            return section_slice0_mapping_.at( brep_component_type )
+                .in2out( section_component );
         }
```
francoisbonneau, line 120: are you sure about that? How do you get the correct mapping?

francoisbonneau (approved):

PR #854, opened 2023-10-10
Allow silent invocation of repair_polygon_orientations(..)

yo35:
To avoid polluting the log when `repair_polygon_orientations(..)` is invoked frequently.

### DIFF
--- a/src/geode/mesh/helpers/repair_polygon_orientations.cpp
+++ b/src/geode/mesh/helpers/repair_polygon_orientations.cpp
@@ -329,7 +329,7 @@ namespace geode
         {
             builder.edges_builder().delete_isolated_edges();
         }
-        Logger::info( "Repair polygons orientations: ",
+        Logger::debug( "Repair polygons orientations: ",
             polygons_to_reorient.size(), " polygons reoriented" );
     }
 


src/geode/mesh/helpers/repair_polygon_orientations.cpp
```diff
@@ -329,17 +330,20 @@ namespace geode
         {
             builder.edges_builder().delete_isolated_edges();
         }
-        Logger::info( "Repair polygons orientations: ",
-            polygons_to_reorient.size(), " polygons reoriented" );
+        if( verbose )
+        {
+            Logger::info( "Repair polygons orientations: ",
```
BotellaA, line 335: Should we add a verbose mode or set the message into a Logger::debug or trace so client code can remove this not very important message?

yo35: Sending the message to `Logger::debug(..)` would be OK in the context of my use-case.

yo35: @BotellaA I've modified the PR to the message to `Logger::debug` instead of adding a verbose flag, as you've proposed.

BotellaA (approved):

BotellaA:
Can you rename your commit message to follow conventional naming? https://www.conventionalcommits.org/en/v1.0.0-beta.4/

Like:
fix(MeshHelpers): update logging output to debug channel in repair_polygons_orientations

PR #862, opened 2023-11-02
fix(Geomtry): add minimum triangle height method into basic object Tr…

panquez:
…iangle

### DIFF
--- a/src/geode/geometry/basic_objects/triangle.cpp
+++ b/src/geode/geometry/basic_objects/triangle.cpp
@@ -262,6 +262,32 @@ namespace geode
         return bbox;
     }
 
+    template < typename PointType, index_t dimension >
+    double GenericTriangle< PointType, dimension >::minimum_height() const
+    {
+        local_index_t max_length_edge_id{ NO_LID };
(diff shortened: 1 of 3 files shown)


src/geode/geometry/basic_objects/triangle.cpp
```diff
@@ -262,6 +262,32 @@ namespace geode
...
+            {
+                max_length_edge_id = edge_id;
+                edge_max_length = edge_length;
+            }
+        }
+        const Point< dimension >& opposite_vertex = vertices_.at(
+            max_length_edge_id == 0 ? 2 : max_length_edge_id - 1 );
+        const Point< dimension >& next_vertex = vertices_.at(
+            max_length_edge_id == 2 ? 0 : max_length_edge_id + 1 );
+        Segment< dimension > longest_edge{ vertices_.at( max_length_edge_id ),
```
BotellaA, line 286:
```suggestion
        const Segment< dimension > longest_edge{ vertices_.at( max_length_edge_id ),
```

BotellaA (approved):

PR #891, opened 2024-01-31
feat(vertices_around_vertex): Added a new method in the meshes API to…

MelchiorSchuh:
… get the vertices around one vertex of the mesh(linked by an edge).

### DIFF
--- a/src/geode/mesh/core/graph.cpp
+++ b/src/geode/mesh/core/graph.cpp
@@ -192,6 +192,21 @@ namespace geode
         return impl_->edges_around_vertex( vertex_id );
     }
 
+    CurveVerticesAroundVertex Graph::vertices_around_vertex(
+        index_t vertex_id ) const
+    {
+        CurveVerticesAroundVertex result;
+        for( const auto& edge_v : this->edges_around_vertex( vertex_id ) )
+        {
+            const auto candidate = this->edge_vertex( edge_v.opposite() );
+            if( absl::c_find( result, candidate ) == result.end() )
+            {
+                result.push_back( candidate );
+            }
+        }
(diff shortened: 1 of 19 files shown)


src/geode/mesh/core/graph.cpp
```diff
@@ -192,6 +192,21 @@ namespace geode
         return impl_->edges_around_vertex( vertex_id );
     }
 
+    CurveVerticesAroundVertex Graph::vertices_around_vertex(
+        index_t vertex_id ) const
+    {
+        CurveVerticesAroundVertex result;
+        for( const auto& edge_v : this->edges_around_vertex( vertex_id ) )
+        {
+            auto candidate = this->edge_vertex( edge_v.opposite() );
```
BotellaA, line 201:
```suggestion
            const auto candidate = this->edge_vertex( edge_v.opposite() );
```

src/geode/mesh/core/surface_mesh.cpp
```diff
@@ -228,6 +228,29 @@ namespace geode
...
         }
 
+        SurfaceVerticesAroundVertex vertices_around_vertex(
+            const SurfaceMesh< dimension >& mesh, index_t vertex_id ) const
+        {
+            SurfaceVerticesAroundVertex result;
+            for( const auto& poly_vertex :
+                mesh.polygons_around_vertex( vertex_id ) )
+            {
+                auto candidate = mesh.polygon_vertex(
```
BotellaA, line 238:
```suggestion
                const auto next_candidate = mesh.polygon_vertex(
```
And rename the next candidate variable

BotellaA (approved):

PR #892, opened 2024-02-06
fix(CMake): add a download target to only download all third parties

BotellaA:


cmake/SuperBuild.cmake
```diff
@@ -18,6 +18,8 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+cmake_policy(SET CMP0114 NEW)
```
panquez, line 21: Only for personal culture, what does it stand for?

panquez (approved):

cmake/SuperBuild.cmake
BotellaA: https://cmake.org/cmake/help/latest/policy/CMP0114.html
mainly a change in the behavior of the dependent steps between several external projects

PR #894, opened 2024-02-07
fix(LTO): missing third party

BotellaA:


src/geode/geometry/distance.cpp
```diff
@@ -70,6 +70,7 @@ namespace
     {
         const auto p =
             ( point_to_v0_length + point_to_v1_length + segment_length ) / 2;
+        if( p - point_to_v0_length <= geode::global_epsilon )
```
BotellaA, line 73: This is a WTF bug!!!

panquez: When had this line been removed? If the bug was not introduced when I pushed the code... :o

panquez (approved):

PR #899, opened 2024-02-26
fix(ModelHelpers): memory issue while converting a model to a mesh

BotellaA:


panquez (approved):

MelchiorSchuh (approved):

MelchiorSchuh (commented):
Maybe add a test so that it doesn't happen again ?

francoisbonneau (approved):

PR #900, opened 2024-02-27
fix(CMake): wrong RPATH for executables after install

yo35:
When creating an executable with `add_geode_binary`, if this executable links a lib created in the same CMake project AND is configured to be installed to a custom directory, it does not work after install.

Example with the attached reproducer:
```
cmake .. -DCMAKE_INSTALL_PREFIX=/path/to/install/dir -DCMAKE_PREFIX_PATH=/path/to/geode
make
./bin/reproducer    # OK
make install
/path/to/install/dir/bin/reproducer    # error while loading shared libraries...
```

Checking the RUNPATH header in the binaries explains this behavior:
```
objdump -x ./bin/reproducer | grep RUNPATH
# RUNPATH    $ORIGIN/../lib:
objdump -x /path/to/install/dir/bin/reproducer | grep RUNPATH
# RUNPATH    /../lib
```
-> the 2nd RUNPATH is invalid.

### DIFF
--- a/cmake/CppTargets.cmake
+++ b/cmake/CppTargets.cmake
@@ -200,7 +200,7 @@ function(_add_geode_executable exe_path folder_name)
     set_target_properties(${target_name}
         PROPERTIES
             FOLDER ${folder_name}
-            INSTALL_RPATH "${OS_RPATH}/../${CMAKE_INSTALL_LIBDIR}"
+            INSTALL_RPATH "$ORIGIN/../${CMAKE_INSTALL_LIBDIR}"
     )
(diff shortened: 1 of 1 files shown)


yo35:
[reproducer-wrong-rpath.zip](https://github.com/Geode-solutions/OpenGeode/files/14417600/reproducer-wrong-rpath.zip)

BotellaA (approved):

BotellaA:
It will be included in the next release! Thanks for your feedback!!

yo35:
Thanks @BotellaA

PR #901, opened 2024-02-27
Feat/export geos

francoisbonneau:


src/geode/mesh/helpers/detail/element_identifier.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include <geode/mesh/core/solid_mesh.h>
+
+namespace geode
+{
+    namespace detail
+    {
+        bool opengeode_mesh_api solid_polyhedron_is_a_tetrahedron(
+            const SolidMesh3D &solid, index_t polyhedron_id )
+        {
+            if( solid.nb_polyhedron_vertices( polyhedron_id ) == 4 )
```
BotellaA, line 35: Return directly the test. No need for an if block

BotellaA (approved):

src/geode/mesh/helpers/detail/element_identifier.cpp
```diff
@@ -0,0 +1,98 @@
...
+            {
+                return false;
+            }
+            for( const auto &facet_vertices : facets_vertices )
+            {
+                if( facet_vertices.size() != 3 )
+                {
+                    return false;
+                }
+            }
```
panquez, line 94: Not sure of that. A prism has only 5 facets, and belong them 3 facets have 4 vertices et the 2 others has 3 vertices

MelchiorSchuh (approved):

src/geode/mesh/helpers/detail/element_identifier.cpp
MelchiorSchuh: check if the facets have only 3 or 4 vertices then ? Is it enough ?

PR #910, opened 2024-03-26
fix(Optional): better use of RVO

BotellaA:


panquez (approved):

src/geode/geometry/basic_objects/triangle.cpp
```diff
@@ -47,13 +47,13 @@ namespace
...
         {
-            PivotNormalResult result;
+            absl::optional< PivotNormalResult > result;
             for( const auto pivot : geode::LRange{ 3 } )
             {
                 const auto next = pivot + 1 == 3 ? 0 : pivot + 1;
                 const geode::Vector3D edge{ points[pivot], points[next] };
-                result.lengths[pivot] = edge.length();
-                const auto edge0 = edge / result.lengths[pivot];
+                result->lengths[pivot] = edge.length();
```
MelchiorSchuh, line 55: This works without initializing the object inside the optional ?

src/geode/mesh/builder/surface_mesh_builder.cpp
```diff
@@ -155,9 +155,11 @@ namespace
         for( const auto e :
             geode::LRange{ surface.nb_polygon_edges( polygon_adj_id ) } )
         {
-            const geode::PolygonEdge adj_edge{ polygon_adj_id, e };
-            const auto adj_v0 = surface.polygon_vertex( adj_edge );
-            const auto adj_v1 = surface.polygon_edge_vertex( adj_edge, 1 );
+            absl::optional< geode::PolygonEdge > adj_edge{ absl::in_place,
```
MelchiorSchuh, line 158: Is the in_place necessary ? I was under the impression that it was done directly with the {}

MelchiorSchuh (approved):

src/geode/geometry/basic_objects/triangle.cpp
BotellaA: It is initialized at optional building.

src/geode/mesh/builder/surface_mesh_builder.cpp
BotellaA: the {} only works if the parameter is the same type as the type inside the optional. If you want to use other parameters to be forwarded to the constructor of the inside type, you need in_place

PR #911, opened 2024-03-26
feat(SolidMesh): add polyhedron_facet_area

MelchiorSchuh:
Added a method on SolidMesh to compute the area of a polyhedron facet. Implementation similar to SurfaceMesh3D::polygon_area.


yo35:
Thanks guys.

PR #921, opened 2024-04-29
fix(SurfaceComponentMeshEdges): assure that component edges does not …

BenPinet:
…return surface edges with identical unique vertices


src/geode/model/helpers/component_mesh_edges.cpp
```diff
@@ -161,11 +161,31 @@ namespace geode
...
                 {
-                    for( auto& polygon_vertex :
-                        mesh.polygons_from_edge_vertices( pair ) )
+                    if( auto edge = mesh.polygon_edge_from_vertices(
+                            pair[0], pair[1] ) )
+
+                    {
+                        edges[surface.id()].emplace_back(
+                            std::move( edge.value() ) );
+                        if( auto adj =
```
BotellaA, line 170: Remove the if blocks. We only want one edge per vertex pair

BotellaA (changes requested):

BotellaA (approved):

PR #929, opened 2024-05-29
fix(Vector): Added Vector to genericable and interpolable attributes.

MelchiorSchuh:


include/geode/geometry/vector.h
```diff
@@ -132,4 +132,51 @@ namespace geode
...
+                "incorrect item value" );
+            return static_cast< float >( vector.value( item ) );
+        }
+
+        static bool is_genericable()
+        {
+            return true;
+        }
+        static local_index_t nb_items()
+
```
BotellaA, line 177:
```suggestion

        static local_index_t nb_items()
```

BotellaA (approved):

PR #933, opened 2024-06-04
feat(BRepHelper): add ray tracing based test for point in block.

francoisbonneau:


include/geode/model/helpers/ray_tracing.h
```diff
@@ -47,4 +47,10 @@ namespace geode
             const BRep& brep,
             const Block3D& block );
 
+    bool opengeode_model_api test_point_inside_block(
```
BotellaA, line 50:
```suggestion
    bool opengeode_model_api is_point_inside_block(
```

include/geode/model/helpers/ray_tracing.h
```diff
@@ -47,4 +47,10 @@ namespace geode
...
+        const BRep& brep, const Block3D& block, const Point3D& point );
+
+    absl::optional< uuid > opengeode_model_api block_containing_point(
```
BotellaA, line 53: What happens if the point is on a surface or on a line? How many uuid should we return?

src/geode/model/helpers/ray_tracing.cpp
```diff
@@ -32,6 +32,55 @@
...
+        const geode::Block3D& block )
+    {
+        geode::index_t nb_intersections{ 0 };
+        const auto tracing =
+            find_intersections_with_boundaries( ray, brep, block );
+        for( const auto& intersections : tracing )
+        {
+            for( const auto& intersection : intersections.second )
+            {
+                if( intersection.position != geode::Position::inside )
```
BotellaA, line 65: If intersection point is on surface triangle edge or vertex, what is the expected behavior?

include/geode/model/helpers/ray_tracing.h
francoisbonneau: if the point is on the boundary it is not in the block. 
Here the function return the block containing the point. 
Should we do a more generic function?

src/geode/model/helpers/ray_tracing.cpp
francoisbonneau: Any advise? the problem is that if the intersection is on a vertex or an edge it is ambiguous in this context.

francoisbonneau: maybe a solution is to slightly change the direction of the ray untill we get a non ambigous response. the risk of infinit loop looks small don't you think?. 
we can either  chose the direction randomly or rotate the direction by a chosen increment.

BotellaA (approved):

PR #937, opened 2024-06-14
fix(ModelBuilder): set point coordinate from unique vertex

BotellaA:


MelchiorSchuh (approved):
First PR where model meshes are created/modified directly from the model builder itself !

PR #955, opened 2024-07-04
fix(Optional): use std::optional

BenPinet:
BREAKING CHANGE: absl::optional has been replaced by std::optional


bindings/python/src/common.h
```diff
@@ -28,8 +28,8 @@
 #include <absl/container/fixed_array.h>
 #include <absl/container/inlined_vector.h>
 #include <absl/strings/string_view.h>
-#include <absl/types/optional.h>
 #include <absl/types/span.h>
+#include <optional>
```
BotellaA, line 32: standard include should be first

include/geode/mesh/core/surface_edges.h
```diff
@@ -23,8 +23,8 @@
 
 #pragma once
 
-#include <absl/types/optional.h>
 #include <absl/types/span.h>
+#include <optional>
```
BotellaA, line 27: here also

BotellaA (approved):

upgrade-guide.md
```diff
@@ -4,12 +4,20 @@
 
 ### Motivations
 
-TODO.....
+
+
 
 ### Breaking Changes
 
+- **Optional**: Replace abls::optional by is equivalent in C++17 std::optional.
```
BotellaA, line 12:
```suggestion
- **Optional**: Replace absl::optional by is equivalent in C++17 std::optional.
```

PR #961, opened 2024-07-05
fix(explicit keyword): Added explicit keyword on several constructors…

MelchiorSchuh:
…, including Point<dim>, Vector<dim>, etc.

BREAKING CHANGE: Most single-parameter constructors are now tagged as explicit, removing the possibility to implicitely construct the objects.


BotellaA (approved):

BotellaA:
@MelchiorSchuh think about the upgrade guide too ;)

PR #973, opened 2024-07-11
fix(Private): rename private folders to internal

BenPinet:
BREAKING CHANGE: change namespace detail in private folders to internal namespace


cmake/CppTargets.cmake
```diff
@@ -257,6 +257,6 @@ function(add_geode_test)
         )
     endif()
     if(USE_BENCHMARK)
-        target_compile_definitions(${target_name} PRIVATE OPENGEODE_BENCHMARK)
+        target_compile_definitions(${target_name} INTERNAL OPENGEODE_BENCHMARK)
```
BotellaA, line 260: nope, this PUBLIC or PRIVATE is from CMake

BotellaA (approved):

PR #977, opened 2024-07-12
fix(TetrahedronSetPoint): replace index_t with local_index_t 

BenPinet:
BREAKING CHANGE: Change GenericTetrahedron set_point parameter from index_t to local_index_t


BotellaA (approved):
maybe check also the other basic object like Edge, Triangle

BenoitTHEBAULT (approved):

PR #992, opened 2024-07-26
fix(ABI): abi fix

BenoitTHEBAULT:


BotellaA (changes requested):
No, this is not possible

PR #997, opened 2024-07-30
feat(transfer_metadata): Added functions to transfer metadata and col…

MelchiorSchuh:
…lections from one model to another with mappings.


include/geode/mesh/core/grid.hpp
```diff
@@ -175,7 +175,7 @@ namespace geode
                 { 2, 6 }, { 3, 7 } } };
 
         static constexpr std::array< std::array< geode::local_index_t, 4 >, 6 >
-            CELL_FACET_VERTICES_3D{ { { 0, 4, 6, 2 }, { 1, 5, 7, 3 },
+            CELL_FACET_VERTICES_3D{ { { 0, 2, 6, 4 }, { 1, 5, 7, 3 },
```
BotellaA, line 178: sure about this change?

src/geode/model/representation/core/detail/transfer_collections.cpp
```diff
@@ -0,0 +1,378 @@
...
+namespace
+{
+    template < typename Model >
+    void transfer_corner_collections( const Model& old_model,
+        const Model& new_model,
+        typename Model::Builder& new_model_builder,
+        const geode::ModelGenericMapping& component_mapping )
+    {
+        if( !component_mapping.has_mapping_type(
+                geode::Corner3D::component_type_static() ) )
```
BotellaA, line 49:
```suggestion
                geode::Corner<Model::dim>::component_type_static() ) )
```
Maybe better like this? And same for the others?

src/geode/model/representation/core/detail/transfer_collections.cpp
```diff
@@ -0,0 +1,378 @@
...
+            transfer_surface_collections(
+                old_brep, new_brep, new_brep_builder, component_mapping );
+            transfer_block_collections(
+                old_brep, new_brep, new_brep_builder, component_mapping );
+            if( !component_mapping.has_mapping_type(
+                    geode::Surface3D::component_type_static() ) )
+            {
+                return;
+            }
+            const auto& surfaces_mapping =
```
BotellaA, line 276: add also a function for this, even if it is used once, so the transfer_brep_collections() has only one level of abstraction

BotellaA (approved):

BotellaA:
Could you use this helper in the Model copy/clone?

include/geode/mesh/core/grid.hpp
MelchiorSchuh: Yes, this was a bug: the facet was not oriented like the others

PR #998, opened 2024-07-30
fix(Geoemtry): fix segment_triangle_intersection tolerance for findin…

panquez:
…g intersections in limit cases


src/geode/geometry/intersection.cpp
```diff
@@ -367,17 +372,19 @@ namespace geode
         }
 
         const Vector3D diff{ vertices[0], line.origin() };
-        const auto DdQxE2 = sign * line.direction().dot( diff.cross( edge2 ) );
-        if( DdQxE2 >= 0 )
+        const auto d_dot_q_cross_e2 =
+            sign * line.direction().dot( diff.cross( edge2 ) );
+        if( d_dot_q_cross_e2 >= 0 )
```
BotellaA, line 377: -GLOBAL_EPSILON ?

src/geode/geometry/intersection.cpp
```diff
@@ -367,17 +372,19 @@ namespace geode
...
         {
-            const auto DdE1xQ =
+            const auto d_dot_e1_cross_q =
                 sign * line.direction().dot( edge1.cross( diff ) );
-            if( DdE1xQ >= 0 && DdQxE2 + DdE1xQ <= DdN )
+            if( d_dot_e1_cross_q >= 0
```
BotellaA, line 381: and here?

BotellaA (approved):

PR #1003, opened 2024-08-02
fix(Bounding Box): add method to get the largest axe of the BBox

panquez:
Do you see an interest to this method?


include/geode/geometry/bounding_box.hpp
```diff
@@ -103,6 +103,8 @@ namespace geode
 
         double smallest_length() const;
 
+        local_index_t largest_dimension() const;
```
panquez, line 106: I wonder if the return should be a tuple with local_index_t and a double

BenoitTHEBAULT (approved):

francoisbonneau (approved):

PR #1006, opened 2024-08-13
feat(ProgressLogger): allow to change the refresh interval

yo35:
The default 1 sec interval may result in verbose logs in case of long jobs.


BotellaA:
Adding an argument to the constructor is breaking the API. See #1011 for another proposition

PR #1011, opened 2024-09-03
fix(ProgressLogger): add refresh interval setter

BotellaA:
@yo35 This is another proposition from your PR. Is it ok for you?


panquez (approved):

yo35:
@BotellaA Thanks Arnaud. This solution is OK as well for our use-case.

PR #1013, opened 2024-09-06
feat(Timer): provide access to duration in raw format

yo35:
Having access to timer duration as `absl::Duration` (instead of just as a human-readable string) is necessary to be able to perform arithmetic operations (e.g. adding timer durations, computing min/max/average durations over several runs, etc.).


BotellaA (approved):

yo35:
Thanks @BotellaA

PR #1017, opened 2024-09-12
fix(BRep_into_solid_conversion): check if all unique_vertices are contained in a block

BenoitTHEBAULT:


panquez (approved):

src/geode/model/helpers/convert_to_mesh.cpp
```diff
@@ -429,6 +429,23 @@ namespace geode
...
+            for( const auto& component :
+                brep.component_mesh_vertices( unique_vertex ) )
+            {
+                if( component.component_id.type()
+                    == geode::Block3D::component_type_static() )
+                {
+                    is_in_block = true;
+                    break;
+                }
+            }
```
BotellaA, line 445: you can use has_component_mesh_vertices(unique_vertex,geode::Block3D::component_type_static() ), it is simpler

BotellaA (approved):

PR #1024, opened 2024-10-02
feat(Geometry): Adding project_point function to Point class and least_meaningful_axis function to Vector class.

BenoitTHEBAULT:


include/geode/geometry/point.hpp
```diff
@@ -174,6 +174,27 @@ namespace geode
             return oss.str();
         }
 
+        [[nodiscard]] Point< dimension - 1 > project_point(
+            geode::local_index_t axis_to_remove ) const
+        {
+            OPENGEODE_ASSERT( axis_to_remove < dimension && axis_to_remove >= 0
+                                  && dimension > 1,
```
panquez, line 181: Dim > 1 useless check done just after

include/geode/geometry/point.hpp
```diff
@@ -174,6 +174,27 @@ namespace geode
...
+                "[Point] Invalid axis to remove" );
+            OPENGEODE_ASSERT(
+                dimension > 1, "[Point] Invalid dimension to reduce" );
+            Point< dimension - 1 > projected_point;
+            geode::index_t dim{ 0 };
+            for( const auto i : LRange{ dimension } )
+            {
+                if( i != axis_to_remove )
+                {
+                    projected_point.set_value( dim, this->value( i ) );
```
panquez, line 191: By doing so you may return primitives. We should choose which algorithm to do if we want to remove coord 1 : either 0,2 or 2,0 !

panquez: We already had this kind of issues in rasterize functions

include/geode/geometry/vector.hpp
```diff
@@ -130,6 +130,74 @@ namespace geode
...
+
+        [[nodiscard]] geode::local_index_t least_meaningful_axis() const
+        {
+            if( dimension == 1 )
+            {
+                return 0;
+            }
+            if( dimension == 2 )
+            {
+                return std::abs( this->value( 0 ) )
```
panquez, line 142: Use std::fabs

include/geode/geometry/vector.hpp
```diff
@@ -130,6 +130,36 @@ namespace geode
...
                     - this->value( 1 ) * other.value( 0 ) } };
         }
+
+        [[nodiscard]] geode::local_index_t least_meaningful_axis() const
+        {
+            geode::local_index_t axis{ 0 };
+            double max{ 0 };
+            for( const auto i : LRange{ dimension } )
+            {
+                if( std::fabs( this->value( i ) ) > max )
```
BotellaA, line 140: Do std::fabs( this->value( i ) ) only once

include/geode/geometry/vector.hpp
```diff
@@ -130,6 +130,36 @@ namespace geode
...
+                    axis = i;
+                }
+            }
+            return axis;
+        }
+
+        [[nodiscard]] geode::local_index_t most_meaningful_axis() const
+        {
+            geode::local_index_t axis{ 0 };
+            double min{ 0 };
```
BotellaA, line 152: no, it should be the max double value

include/geode/geometry/point.hpp
BotellaA: I would say 0, 2 to keep the same order, we could implement a reorder/permute method if needed

BotellaA: How do you want to choose an order if not keeping the same ?

include/geode/geometry/point.hpp
```diff
@@ -174,6 +174,26 @@ namespace geode
             return oss.str();
         }
 
+        [[nodiscard]] Point< dimension - 1 > project_point(
```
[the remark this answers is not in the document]
BotellaA: https://github.com/Geode-solutions/OpenGeode/blob/master/include/geode/geometry/projection.hpp#L76

include/geode/geometry/vector.hpp
```diff
@@ -130,6 +130,37 @@ namespace geode
...
+                    axis = i;
+                }
+            }
+            return axis;
+        }
+
+        [[nodiscard]] geode::local_index_t least_meaningful_axis() const
+        {
+            geode::local_index_t axis{ 0 };
+            double min{ std::numeric_limits< double >::max() };
```
BotellaA, line 153:
```suggestion
            auto min = std::numeric_limits< double >::max();
```
double is already force in the template

include/geode/geometry/vector.hpp
```diff
@@ -130,6 +130,37 @@ namespace geode
...
+            for( const auto i : LRange{ dimension } )
+            {
+                if( std::fabs( this->value( i ) ) < min )
```
BotellaA, line 156: do not compute the fabs twice

BotellaA (approved):

PR #1026, opened 2024-10-11
fix(convert_grid_to_solid): Added possibility to densify pattern of c…

MelchiorSchuh:
…ell conversion to tetrahedra.


src/geode/mesh/helpers/convert_solid_mesh.cpp
```diff
@@ -66,74 +155,71 @@ namespace
                 for( const auto i :
                     geode::Range{ grid.nb_cells_in_direction( 0 ) } )
                 {
-                    const auto cell_vertices =
-                        grid.cell_vertices( { i, j, k } );
                     const auto cell = grid.cell_index( { i, j, k } );
-                    std::array< geode::index_t, 8 > cell_mesh_vertices;
-                    for( const auto v : geode::LIndices{ cell_mesh_vertices } )
+                    if( absl::c_find( cells_to_densify, cell )
```
BotellaA, line 159: To check if this find cost a lot. If yes, you could create a vector<bool>, iterate on cells_to_densify and tag them. I think it would be faster

BotellaA (approved):

PR #1032, opened 2024-10-21
fix(BasicObjects): add is_degenerated

BotellaA:


src/geode/geometry/basic_objects/triangle.cpp
```diff
@@ -269,6 +270,28 @@ namespace geode
...
+            const Segment< dimension > edge{ point0, point1 };
+            if( edge.is_degenerated() )
+            {
+                return true;
+            }
+            const auto last_vertex = next_vertex == 2 ? 0 : next_vertex + 1;
+            const Point< dimension >& point2 = vertices_.at( last_vertex );
+            return point_line_distance(
+                       point2, InfiniteLine< dimension >{ edge } )
+                   <= GLOBAL_EPSILON;
```
panquez, line 290: Not a return her, but if <= GLOBAL_EPSILON return true

src/geode/geometry/basic_objects/triangle.cpp
```diff
@@ -269,6 +270,28 @@ namespace geode
...
+        }
+        return true;
```
panquez, line 292: return false

src/geode/geometry/basic_objects/tetrahedron.cpp
```diff
@@ -79,6 +82,35 @@ namespace geode
...
+            {
+                if( vertex3 == vertex0 || vertex3 == vertex1
+                    || vertex3 == vertex2 )
+                {
+                    continue;
+                }
+                const Point3D& point3 = vertices_.at( vertex3 );
+                return std::get< 0 >(
+                           point_plane_distance( point3, Plane{ triangle } ) )
+                       <= GLOBAL_EPSILON;
```
panquez, line 108: not a return here. but if <= GLOBAL_EPSILON then return true, else go for the next tetrahedron_facet

src/geode/geometry/basic_objects/tetrahedron.cpp
```diff
@@ -79,6 +82,35 @@ namespace geode
...
+            }
+        }
+        return true;
```
panquez, line 111: return false, if it is ok for all 4 facets

src/geode/geometry/basic_objects/tetrahedron.cpp
panquez: Or you do not have to iterates in the 4 facets, since you will directly have an answer on the first tested facet

panquez (changes requested):

src/geode/geometry/basic_objects/tetrahedron.cpp
BotellaA: Might be enough... If it is degen then the simplex is degen, is not the distance once should be enough to determine since we are in a simplex. What do you think?

panquez: So remove the main for loop?

panquez (approved):

PR #1035, opened 2024-10-24
fix(Cached Value): skip serialize to reduce file size

BotellaA:


BotellaA:
> Is this compatible with old files? Is it still possible to deserialize them?

Yes, the first deserializer should still works. Only new files will not save the data.

panquez (approved):

PR #1038, opened 2024-10-28
fix(gradient computation): Added functions to compute gradient giving…

MelchiorSchuh:
… no value vertices instead of nan, and gradient shape functions for grids.


bindings/python/requirements.txt
BotellaA, line 1: this file should not exit anymore, pull from next maybe?

include/geode/mesh/helpers/generic_solid_accessor.hpp
```diff
@@ -45,14 +45,40 @@ namespace geode
     public:
         using ElementVertex = PolyhedronVertex;
         using ElementVertices = PolyhedronVertices;
-        using ElementFacet = PolyhedronFacet;
         using ElementFacetVertices = PolyhedronFacetVertices;
 
+        struct ElementFacet
```
BotellaA, line 50: why not using ElementFacet = PolyhedronFacet ?

BotellaA (approved):

PR #1047, opened 2024-11-21
fix(Exception): add basic object exceptions

BotellaA:


src/geode/basic/assert.cpp
BotellaA, line 1: All abseil exceptions seem to be directly implemented as std::exceptions

include/geode/geometry/basic_objects/segment.hpp
```diff
@@ -102,4 +103,41 @@ namespace geode
...
+// NOLINTNEXTLINE
+#define OPENGEODE_SEGMENT_EXCEPTION( dimension, condition, segment, ... )      \
+    if( ABSL_PREDICT_FALSE( !( condition ) ) )                                 \
+        throw geode::OpenGeodeSegmentException< dimension >                    \
+        {                                                                      \
+            segment, __VA_ARGS__                                               \
+        }
+
+// NOLINTNEXTLINE
+#define OPENGEODE_SEGMENT_EXCEPTION1D( condition, segment, ... )               \
```
panquez, line 134: Why not put the 1D/2D/3D right after "SEGMENT"?

src/geode/basic/assert.cpp
BotellaA: There is no absl::exception

MelchiorSchuh (approved):

panquez (approved):

PR #1048, opened 2024-11-21
fix(Return): remove [[nodiscard]] specifier on find_or_create_attribute(..)

yo35:
It may be relevant to ignore the pointer returned by `find_or_create_attribute(..)` if the caller just wants to create the attribute (or to ensure it exists).

### DIFF
--- a/include/geode/basic/attribute_manager.hpp
+++ b/include/geode/basic/attribute_manager.hpp
@@ -103,10 +103,10 @@ namespace geode
          * @exception OpenGeodeException if the Attribute replacement failed
          */
         template < template < typename > class Attribute, typename T >
-        [[nodiscard]] std::shared_ptr< Attribute< T > >
-            find_or_create_attribute( std::string_view name,
-                T default_value,
-                AttributeProperties properties )
+        std::shared_ptr< Attribute< T > > find_or_create_attribute(
+            std::string_view name,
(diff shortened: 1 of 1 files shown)


include/geode/basic/attribute_manager.hpp
```diff
@@ -127,8 +127,8 @@ namespace geode
         }
 
         template < template < typename > class Attribute, typename T >
```
BotellaA, line 129: This is no likely to be accepted. Why do you need that?

You can use a workaround using either geode_unused function of [[maybe_unused]] keyword.

yo35:
I need that to create an attribute, but I don't need the pointer returned by the method. And there could be several reasons why I don't need it:
- because the attribute that is created is constant,
- because the creation and the modification of the attribute in the calling lib is not performed at the same place (e.g. the former in a sequential segment, the latter in a parallel segment),
- etc...

More generally, in my opinion, ``[[nodiscard]]`` should not be applied to a method whose purpose is to change the state of one of its argument (i.e. to perform a side-effect operation).

include/geode/basic/attribute_manager.hpp
yo35: Hi @BotellaA . Answer in main thread.

PR #1052, opened 2024-11-28
fix(Triangulate): change algorithm for ear clipping

BotellaA:


src/geode/mesh/helpers/convert_surface_mesh.cpp
```diff
@@ -384,20 +414,50 @@ namespace geode
...
                 for( const auto e : LRange{ nb_vertices } )
                 {
-                    adjacents[e] = surface.polygon_adjacent_edge( { p, e } );
+                    if( const auto adj =
+                            surface.polygon_adjacent_edge( { p, e } ) )
+                    {
+                        adjacents.emplace(
+                            Edge{ vertices[e],
+                                vertices[e + 1 == nb_vertices ? 0 : e + 1] },
+                            adj.value() );
```
panquez, line 428: are there always a value?

BotellaA: No but it is protected inside an if block

panquez (approved):

PR #1054, opened 2024-11-29
fix(CMake): fix library paths for test executables

yo35:
- Avoid duplicated folders
- Use target property INTERFACE_LINK_LIBRARIES instead of LINK_LIBRARIES, as the latter is not exported in PackageConfig.cmake files


yo35:
@BotellaA As we discussed, this does not solve the problem of dynamic libraries registered as private dependencies, nor the fact that `$ENV{LD_LIBRARY_PATH}` is evaluated during the configuration phase, not when running the tests.

I mentioned that this is completely **untested** under Windows. ;-)

PR #1055, opened 2024-11-29
fix(CMake): fix library paths for test executables

BotellaA:
- Avoid duplicated folders
- Use target property INTERFACE_LINK_LIBRARIES instead of LINK_LIBRARIES, as the latter is not exported in PackageConfig.cmake files

@yo35  I moved the PR internally to check with Github actions


panquez (approved):

yo35:
Cannot investigate by myself why this fails on Windows, sorry.

The point aimed to be fixed by this PR is not blocking on my side, so feel free to discard this PR if you don't want to dig further.

PR #1059, opened 2024-12-03
 fix(AABBTree): allow async search

BotellaA:


tests/geometry/test-aabb.cpp
```diff
@@ -411,6 +417,8 @@ void test_self_intersections()
     eval_intersection.included_box_.clear();
     aabb.compute_self_element_bbox_intersections( eval_intersection );
 
+    DEBUG( eval_intersection.included_box_.size() );
+    DEBUG( nb_boxes * nb_boxes );
```
panquez, line 421: to delete?

panquez (approved):

PR #1064, opened 2024-12-18
fix(TetrahedralSolid): adding function to get polyhedron facet triangle (with geometrical points).

BenoitTHEBAULT:


panquez (changes requested):
This function already exists and it is called `triangle()`

BotellaA (dismissed):

BotellaA:
> This function already exists and it is called `triangle()`

True! I missed this one

PR #1080, opened 2025-01-30
fix(Basic): missing include

yo35:


BotellaA (approved):

yo35:
This aims to fix the following error, encountered with GCC 13.3:

```
In file included from ...:
.../geode/include/geode/basic/types.hpp:52:71: error: return type ‘std::string’ {aka ‘class std::basic_string<char>’} is incomplete
   52 |     [[nodiscard]] inline std::string to_string( std::string_view view )
      |                                                                       ^
.../geode/include/geode/basic/types.hpp: In function ‘void geode::to_string(std::string_view)’:
.../geode/include/geode/basic/types.hpp:54:49: error: invalid use of incomplete type ‘std::string’ {aka ‘class std::basic_string<char>’}
   54 |         return static_cast< std::string >( view );
      |                                                 ^

```

PR #1083, opened 2025-02-12
fix(Geometry): now unique points from ColocatedInfo have distances wi…

panquez:
…der than given epsilon from each other

### DIFF
--- a/src/geode/geometry/nn_search.cpp
+++ b/src/geode/geometry/nn_search.cpp
@@ -23,6 +23,7 @@
 
 #include <geode/geometry/nn_search.hpp>
 
+#include <mutex>
 #include <numeric>
 
 #include <absl/algorithm/container.h>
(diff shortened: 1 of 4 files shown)


src/geode/geometry/nn_search.cpp
```diff
@@ -190,14 +210,27 @@ namespace geode
...
         async::parallel_for( async::irange( index_t{ 0 }, nb_points() ),
-            [&epsilon, &mapping, this]( index_t p ) {
-                if( mapping[p] == p )
+            [&epsilon, &mapping, &mutex, this]( index_t p ) {
+                if( mapping[p] != p )
+                {
+                    return;
+                }
+                const auto vertices = radius_neighbors( point( p ), epsilon );
+                auto min_index = find_min_unmapped_element( vertices, mapping );
```
BotellaA, line 221: not sure to see the difference with min_element(vertices)

BotellaA (approved):

src/geode/geometry/nn_search.cpp
panquez: considering only element == mapping[element]
it is the crucial change of this PR

PR #1102, opened 2025-03-07
fix(Identifier): set default name as uuid

panquez:
What do you think about this change?

### DIFF
--- a/src/geode/basic/identifier.cpp
+++ b/src/geode/basic/identifier.cpp
@@ -34,7 +34,13 @@ namespace geode
     class Identifier::Impl
     {
     public:
-        Impl() = default;
+        Impl()
+        {
+            if( name_ == DEFAULT_NAME )
+            {
+                name_ = id_.string();
+            }
+        }
 
         const uuid& id() const
         {


src/geode/basic/identifier.cpp
```diff
@@ -34,7 +34,13 @@ namespace geode
     class Identifier::Impl
     {
     public:
-        Impl() = default;
+        Impl()
+        {
+            if( name_ == DEFAULT_NAME )
+            {
```
francoisbonneau, line 40: any motivation for that change?
This name will persist even if the uuid changes... I would not do that.

panquez:
@BotellaA I have done the change proposed in your comment. Please could you update the serialize to consider know an optional?

BotellaA:
> @BotellaA I have done the change proposed in your comment. Please could you update the serialize to consider know an optional?

Done

BotellaA (approved):

BotellaA:
Having an algorithm that checks if there is a name by checking the name value is for sure a really bad design choice!! When need a may to know when a name is set or not.
image

PR #1105, opened 2025-03-11
fix(RegularGrid): Fixed issue on grid origin initialization on old gr…

MelchiorSchuh:
…ids.


src/geode/mesh/core/grid.cpp
```diff
@@ -26,6 +26,7 @@
 #include <absl/container/inlined_vector.h>
 
 #include <geode/basic/bitsery_archive.hpp>
+#include <geode/basic/logger.hpp>
```
BotellaA, line 29: To remove?

BotellaA (approved):

PR #1115, opened 2025-03-27
fix(Basic): add comparison operators for small sets

panquez:


include/geode/basic/small_set.hpp
```diff
@@ -85,6 +85,27 @@ namespace geode
             return container_.at( index );
         }
 
+        auto operator==( const SmallSet& other ) const
```
BotellaA, line 88: It is not really an operator== since it order independent

BotellaA (approved):

PR #1136, opened 2025-04-17
fix(Build): improve build time

BotellaA:


panquez (approved):

src/geode/basic/assert.cpp
```diff
@@ -25,10 +25,32 @@
...
+#include <absl/debugging/symbolize.h>
+
 #include <geode/basic/logger.hpp>
 
 namespace geode
 {
+    std::string OpenGeodeException::stack_trace() const
+    {
+        std::string stack_string;
+        for( auto frame = 0; frame < stack_size_; ++frame )
```
MelchiorSchuh, line 37: Should we do a range iteration instead ?

MelchiorSchuh (approved):

PR #1154, opened 2025-06-20
feat(AABBTree): generic intersection search method

yo35:


BotellaA:
I forgot about it.. Can you implement a unit test?

yo35:
I can implement a dedicated unit test. I can also re-implement `AABBTree::compute_ray_element_bbox_intersection(..)` using the new function, so that it will be covered by the existing test.

PR #1155, opened 2025-06-23
fix(block_bounding_box): Added case to compute block bounding box fro…

MelchiorSchuh:
…m block mesh if it exists.

### DIFF
--- a/include/geode/mesh/helpers/ray_tracing.hpp
+++ b/include/geode/mesh/helpers/ray_tracing.hpp
@@ -82,7 +82,7 @@ namespace geode
         [[nodiscard]] std::optional< EdgeDistance > closest_edge() const;
 
         [[nodiscard]] std::optional< absl::FixedArray< EdgeDistance > >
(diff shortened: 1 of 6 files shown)


include/geode/mesh/helpers/ray_tracing.hpp
```diff
@@ -35,12 +35,63 @@
...
+    public:
+        RayTracing2D( const EdgedCurve2D& mesh, const Ray2D& ray );
+        RayTracing2D(
+            const EdgedCurve2D& mesh, const InfiniteLine2D& infinite_line );
+        RayTracing2D( RayTracing2D&& other ) noexcept;
+        ~RayTracing2D();
+
+        [[nodiscard]] std::optional< EdgeDistance > closest_edge() const;
+
+        [[nodiscard]] std::optional< absl::FixedArray< EdgeDistance > >
```
BotellaA, line 84: this API is no clear... What does size mean?

BotellaA (approved):

PR #1157, opened 2025-06-24
fix(ModelMappings): intermediate refactoring using more templated str…

panquez:
…ucts


src/geode/model/helpers/detail/mappings_merger.cpp
```diff
@@ -250,10 +250,11 @@ namespace geode
             return transferer.transfer();
         }
 
-        ModelMappings merge_mappings(
-            const ModelMappings& mappings1, const ModelMappings& mappings2 )
+        template < typename ModelMappingType >
+        ModelMappingType base_merge_mappings( const ModelMappingType& mappings1,
```
BotellaA: To move in anonymous namespace?

PR #1178, opened 2025-08-18
feat(BoundingBox): add n_volume computation

francoisbonneau:


src/geode/geometry/bounding_box.cpp
```diff
@@ -508,6 +508,17 @@ namespace geode
         return -inner_distance;
     }
 
+    template < index_t dimension >
+    double BoundingBox< dimension >::n_volume() const
+    {
+        double volume{ 1.0 };
+        for( const auto c : geode::LRange{ dimension } )
+        {
+            volume *= ( max_.value( c ) - min_.value( c ) );
```
panquez: You can compute diagonal and multiply all the value of the given vector instead.

PR #1181, opened 2025-09-10
fix(TriangleRasterization): Fixed cases of triangle rasterization whe…

MelchiorSchuh:
…re edges of triangle where > epsilon but not edges of triangle in grid

### DIFF
--- a/src/geode/mesh/helpers/rasterize.cpp
+++ b/src/geode/mesh/helpers/rasterize.cpp
@@ -318,8 +318,15 @@ namespace
         };
         const auto triangle_edges = get_triangle_edges( triangle );
         const auto normal_in_grid = triangle_in_grid.normal();
+        const auto triangle_edges_in_grid =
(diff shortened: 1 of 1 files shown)


src/geode/mesh/helpers/rasterize.cpp
```diff
@@ -318,8 +318,15 @@ namespace
         };
         const auto triangle_edges = get_triangle_edges( triangle );
         const auto normal_in_grid = triangle_in_grid.normal();
+        const auto triangle_edges_in_grid =
+            get_triangle_edges( triangle_in_grid );
         if( !normal_in_grid
             || absl::c_count_if( triangle_edges,
+                   []( const geode::Segment3D& segment ) {
```
BotellaA:
```suggestion
                   []( const auto& segment ) {
```

BotellaA (approved):

PR #1183, opened 2025-09-14
feat(Models): add function to return the BBox of active componements

francoisbonneau:


BotellaA (approved):

panquez (approved):

src/geode/model/representation/core/section.cpp
```diff
@@ -600,4 +600,22 @@ namespace geode
         }
         return internal::meshes_bounding_box< 2 >( corners() );
     }
+
+    BoundingBox2D BRep::active_components_bounding_box() const
```
BotellaA:
```suggestion
    BoundingBox2D Section::active_components_bounding_box() const
```

PR #1184, opened 2025-09-15
fix(Quality): adding function to compute triangle quality, based on angles.

BenoitTHEBAULT:


include/geode/geometry/quality.hpp
```diff
@@ -39,4 +41,8 @@ namespace geode
 
     [[nodiscard]] double opengeode_geometry_api
         tetrahedron_collapse_aspect_ratio( const Tetrahedron& tetra );
+
+    template < index_t dimension >
+    [[nodiscard]] double opengeode_geometry_api triangle_angle_based_quality(
```
BotellaA:
```suggestion
    [[nodiscard]] double triangle_angle_based_quality(
```
Do not export templated functions in the header

panquez (approved):
Some DEBUG to remove?

PR #1187, opened 2025-09-18
fix(conservative_rasterize_segment): Filtered result to avoid adding …

MelchiorSchuh:
…cells outside of scope

### DIFF
--- a/src/geode/mesh/helpers/rasterize.cpp
+++ b/src/geode/mesh/helpers/rasterize.cpp
@@ -84,7 +84,10 @@ namespace
             index[axis0] += increments[axis0];
             painted_cells.push_back( index );
         }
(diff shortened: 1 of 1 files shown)


src/geode/mesh/helpers/rasterize.cpp
```diff
@@ -98,12 +101,17 @@ namespace
         std::array< int, dimension > increments;
         for( const auto i : geode::LRange{ dimension } )
         {
+            if( end[i] == start[i] )
+            {
+                deltas[i] = 0;
+                increments[i] = 0;
+            }
             if( end[i] > start[i] )
```
BotellaA:
```suggestion
            else if( end[i] > start[i] )
```

BotellaA (approved):

PR #1197, opened 2025-10-03
fix(BRep): Avoid using try/catch in BRep bounding_blox computation

MelchiorSchuh:


include/geode/mesh/core/solid_mesh.hpp
```diff
@@ -452,7 +452,8 @@ namespace geode
             const PolyhedronFacet& polyhedron_facet ) const;
 
         /*!
-         * Return the normal of a given PolyhedronFacet.
+         * Return the normal of a given PolyhedronFacet (should be oriented
+         * towards the inside of the polyhedron).
```
panquez: True in our geode convention, but is it always true?

BotellaA (approved):

src/geode/model/representation/core/brep.cpp
```diff
@@ -891,34 +891,36 @@ namespace geode
...
-                "with not meshes Blocks and no Corners, no Lines and no "
-                "Surfaces." );
-        }
+        OPENGEODE_EXCEPTION( bbox_computed,
+            "[BRep::bounding_box] Cannot return the bounding_box of a BRep "
+            "with not meshed Blocks and no Corners, no Lines and no "
+            "Surfaces." );
         OPENGEODE_EXCEPTION( bbox.min() <= bbox.max(),
             "[BRep::bounding_box] Cannot return the "
             "bounding_box of an empty BRep." );
```
panquez: I think these two exceptions are similar

MelchiorSchuh: I was unsure wether to remove one of the to but I agree

include/geode/mesh/core/solid_mesh.hpp
MelchiorSchuh: I can add that it is the geode convention

PR #1201, opened 2025-10-14
fix(Model): better handle BBox of empty meshes

BotellaA:


src/geode/model/representation/core/section.cpp
```diff
@@ -597,19 +597,23 @@ namespace geode
...
             bbox.add_box( box.value() );
+            bbox_computed = true;
         }
         if( const auto box = internal::meshes_bounding_box< 2 >( corners() ) )
         {
             bbox.add_box( box.value() );
+            bbox_computed = true;
         }
-        OPENGEODE_EXCEPTION( bbox.min() <= bbox.max(),
+        OPENGEODE_EXCEPTION( bbox_computed && bbox.min() <= bbox.max(),
```
panquez: I would say that to know if bbox is computed we only need either the bool `bbox_computed`  or the STRICT comparison between bbox.min() and bbox.max().
Checking both is not needed

PR #1225, opened 2026-01-12
fix(SolidMesh): Fixed call to specific functions when called on inval…

MelchiorSchuh:
…id mesh

### DIFF
--- a/src/geode/geometry/basic_objects/triangle.cpp
+++ b/src/geode/geometry/basic_objects/triangle.cpp
@@ -60,9 +60,14 @@ namespace
             }
             const auto edge0 = edge / result->lengths[pivot];
             const auto prev = pivot == 0 ? 2 : pivot - 1;
-            const auto edge1 = geode::Vector3D{ points[pivot], points[prev] };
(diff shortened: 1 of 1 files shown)


src/geode/geometry/basic_objects/triangle.cpp
```diff
@@ -48,34 +48,30 @@ namespace
...
-        try
+        std::optional< PivotNormalResult > result{ std::in_place };
+        for( const auto pivot : geode::LRange{ 3 } )
         {
-            std::optional< PivotNormalResult > result{ std::in_place };
-            for( const auto pivot : geode::LRange{ 3 } )
+            const auto next = pivot + 1 == 3 ? 0 : pivot + 1;
+            const geode::Vector3D edge{ points[pivot], points[next] };
+            result->lengths[pivot] = edge.length();
+            if( edge.length() < geode::GLOBAL_EPSILON )
```
BotellaA:
```suggestion
            if( result->lengths[pivot] < geode::GLOBAL_EPSILON )
```

PR #1226, opened 2026-01-19
perf(Quality): small perf improvements

BotellaA:


src/geode/geometry/quality.cpp
```diff
@@ -233,21 +215,16 @@ namespace geode
...
-        }
+        const Vector3D edge01{ vertices[0], vertices[1] };
+        const Vector3D edge02{ vertices[0], vertices[2] };
+        const Vector3D edge03{ vertices[0], vertices[3] };
+        const Vector3D edge12{ vertices[1], vertices[2] };
+        const Vector3D edge13{ vertices[1], vertices[3] };
+        const Vector3D edge23{ vertices[2], vertices[3] };
+        const auto sq_len =
+            std::max( { edge01.length2(), edge02.length2(), edge03.length2(),
+                edge12.length2(), edge13.length2(), edge23.length2() } );
```
panquez: It changes the equation : here is max instead of sum. Is it intentional?

BotellaA: no

PR #1254, opened 2026-03-27
fix(ExplicitVariability): change many Flat_hash_map container into Bt…

MoulardM:
…ree_map to prevent variability in the results

### DIFF
--- a/include/geode/basic/mapping.hpp
+++ b/include/geode/basic/mapping.hpp
@@ -42,8 +42,8 @@ namespace geode
 
         [[deprecated]] void reserve( index_t capacity )
         {
-            // in2out_.reserve( capacity );
-            // out2in_.reserve( capacity );
+            in2out_.reserve( capacity );
+            out2in_.reserve( capacity );
         }
 
         void clear()
--- a/include/geode/basic/mapping.hpp
+++ b/include/geode/basic/mapping.hpp
@@ -23,7 +23,8 @@
 
 #pragma once
 
-#include <absl/container/flat_hash_map.h>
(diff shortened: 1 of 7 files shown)


include/geode/basic/mapping.hpp
```diff
@@ -39,10 +40,10 @@ namespace geode
         template < typename T >
         using Storage = typename StorageType< T >::Type;
 
-        void reserve( index_t capacity )
+        void reserve( index_t /* capacity */ )
```
panquez: Add [[deprecated]] before the function. Add remove calls of this method in all the codes

include/geode/basic/mapping.hpp
```diff
@@ -225,6 +226,11 @@ namespace geode
             }
             auto& out_map = this->out2in_mapping().at( out );
             const auto itr2 = absl::c_find( out_map, in );
+            DEBUG( itr2 == out_map.end() );
```
panquez: to remove

include/geode/model/mixin/core/component_type.hpp
```diff
@@ -84,6 +84,15 @@ namespace geode
             return type_.get() == other.type_.get() && id_ == other.id_;
         }
 
+        [[nodiscard]] bool operator<( const ComponentID& other ) const
+        {
+            if( type_.get() != other.type_.get() )
+            {
+                return type_.get() < other.type_.get();
+            }
+            return id_.string() < other.id_.string();
```
panquez: to not use .string()

panquez (changes requested):

PR #1260, opened 2026-04-20
fix(PolygonsAroundVertex): handling of non-orientable meshes

MoulardM:


tests/mesh/test-moebius-strip.cpp
```diff
@@ -0,0 +1,105 @@
...
+        "[TEST] Wrong polygons around vertex 99" );
+    OPENGEODE_EXCEPTION( polygons_around_99[3].polygon_id == 7
+                             && polygons_around_99[3].vertex_id == 0,
+        "[TEST] Wrong polygons around vertex 99" );
+}
+
+void test()
+{
+    geode::OpenGeodeMeshLibrary::initialize();
+    geode::Logger::set_level( geode::Logger::LEVEL::info );
```
panquez: This line is not useful since it is already by default set to info level

panquez (approved):

PR #1262, opened 2026-04-21
fix(RayTracing): add less axes-aligned vector to detect point inside …

panquez:
…volume

### DIFF
--- a/src/geode/model/helpers/ray_tracing.cpp
+++ b/src/geode/model/helpers/ray_tracing.cpp
@@ -40,28 +40,41 @@
 
 namespace
 {
-    const std::array< geode::Vector2D, 8 > directions_2D = {
-        { geode::Vector2D{ { 1., 0. } }, geode::Vector2D{ { 0., 1. } },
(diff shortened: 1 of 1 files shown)


BotellaA (approved):

src/geode/model/helpers/ray_tracing.cpp
```diff
@@ -135,12 +152,12 @@ namespace geode
...
     {
     public:
-        Impl( const Section& section ) : section_( section ) {}
+        explicit Impl( const Section& section ) : section_( section ) {}
 
         bool is_point_inside_surface(
             const Point2D& point, const Surface2D& surface )
         {
-            for( const auto& direction : directions_2D )
+            for( const auto& direction : ::directions_2D() )
```
francoisbonneau, line 160: je connais pas ca... ca veut dire quoi?

francoisbonneau (approved):

MelchiorSchuh (approved):

src/geode/model/helpers/ray_tracing.cpp
MelchiorSchuh: :: with nothing in front is to precise that the function is in the anonymous namespace

PR #1264, opened 2026-04-23
feat(Exception): new exception at library scope

BotellaA:
BREAKING CHANGE: EXCEPTION and ASSERT macros are removed


include/geode/basic/pimpl.hpp
```diff
@@ -25,18 +25,16 @@
...
     template < typename T >
     class PImpl
     {
-        OPENGEODE_DISABLE_COPY( PImpl );
-
     public:
         template < typename... Args >
         explicit PImpl( Args &&... );
+        PImpl( const PImpl< T > &other ) = delete;
+        PImpl< T > &operator=( const PImpl< T > &other ) = delete;
```
francoisbonneau, line 37: why did yoiu removed the macro?

include/geode/basic/variable_attribute.hpp
```diff
@@ -471,7 +474,8 @@ namespace geode
             {
                 for( const auto new_index : outs )
                 {
-                    OPENGEODE_EXCEPTION( new_index < nb_elements,
+                    OpenGeodeBasicException::check( new_index < nb_elements,
+                        nullptr, OpenGeodeException::TYPE::data,
```
panquez: Since all OpenGeodeBasicException will be filled with nullptr, is it something good or wrong to define a specific check method only for OpenGeodeBasicException without this parameter?

panquez (approved):

include/geode/basic/pimpl.hpp
BotellaA: there is a circular dependency, not sure how to handle it properly. Not including common.hpp works.

include/geode/basic/variable_attribute.hpp
BotellaA: no simple to do because hte class OpenGeodeBasicException is generated with a macro. So we need to customize the macro or create another one. Also, not sure if in the end they all will be nullptr, maybe we could add data as string or uuid

PR #1271, opened 2026-05-13
fix(Exception): updating throw_lippincott to OpenGeodeException

BotellaA:


include/geode/basic/assert.hpp
```diff
@@ -152,5 +139,36 @@ namespace geode
     /*!
      * Catch all exceptions and rethrow an OpenGeodeException
      */
-    void opengeode_basic_api throw_lippincott();
+    template < typename Exception, typename... Args >
+    [[noreturn]] void throw_lippincott(
```
francoisbonneau, line 143: @BotellaA @panquez how can I throw without changing the type?

francoisbonneau (approved):

PR #1277, opened 2026-05-28
feat(Attribute): attribute filtering on import

yo35:


yo35:
An alternative implementation could be to add an overload of the `import(..)` methods with an additional argument `std::string_view attribute_name`, and to let the caller import attributes one by one. Maybe this is better than what I propose here.

PR #1278, opened 2026-05-29
feat(Attribute): attribute filtering on import

BotellaA:
cc @yo35

### DIFF
--- a/include/geode/basic/attribute_manager.hpp
+++ b/include/geode/basic/attribute_manager.hpp
@@ -23,6 +23,7 @@
 
 #pragma once
 
+#include <functional>
 #include <memory>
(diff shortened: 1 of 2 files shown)


panquez (approved):

yo35:
@BotellaA @panquez As I said in #1277, maybe I can propose a solution more simple, by overloading methods `import(..)` with a additional `std::string_view attribute_name` argument, to allow importing attributes one by one. I think it would lead to a simpler API.

BotellaA:
@yo35 propose your second idea and we will see which one looks simpler. Probably the second one

yo35:
Alternative solution proposed in #1279

PR #1279, opened 2026-05-29
feat(Attribute): import methods for a single attribute

yo35:


src/geode/basic/attribute_manager.cpp
```diff
@@ -310,6 +348,44 @@ namespace geode
             }
         }
 
+        void import( const AttributeManager::Impl &attribute_manager,
```
BotellaA, line 286: could you reuse this code to import all attributes? I think there are code duplication here

yo35: Sharing code for import all vs import one is not that simple, because of exception handling. But factorization is possible between the two import-all and import-one versions.

yo35:
Thanks @BotellaA

PR #1285, opened 2026-06-08
fix(FailedMedialBallTest) : store AABB trees as unique_ptr in ray_tra…

MoulardM:
…cing

### DIFF
--- a/src/geode/model/helpers/ray_tracing.cpp
+++ b/src/geode/model/helpers/ray_tracing.cpp
@@ -302,14 +302,14 @@ namespace geode
         {
             {
                 absl::ReaderMutexLock read_lock{ mutex_ };
-                auto it = aabb_trees_.find( surface.id() );
+                const auto it = aabb_trees_.find( surface.id() );
                 if( it != aabb_trees_.end() )
                 {
                     return *it->second;
                 }
             }
             absl::MutexLock lock{ mutex_ };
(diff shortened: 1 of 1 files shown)


src/geode/model/helpers/ray_tracing.cpp
```diff
@@ -302,20 +302,22 @@ namespace geode
         {
             {
                 absl::ReaderMutexLock read_lock{ mutex_ };
-                if( aabb_trees_.contains( surface.id() ) )
+                auto it = aabb_trees_.find( surface.id() );
```
BotellaA:
```suggestion
                const auto it = aabb_trees_.find( surface.id() );
```

src/geode/model/helpers/ray_tracing.cpp
```diff
@@ -302,20 +302,22 @@ namespace geode
...
                 {
-                    return aabb_trees_.at( surface.id() );
+                    return *it->second;
                 }
             }
             absl::MutexLock lock{ mutex_ };
-            return aabb_trees_
-                .emplace( surface.id(), create_aabb_tree( surface.mesh() ) )
-                .first->second;
+            auto [it, inserted] = aabb_trees_.emplace(
```
BotellaA:
```suggestion
            const auto [it, inserted] = aabb_trees_.emplace(
```

BotellaA (approved):

PR #1293, opened 2026-06-23
fix(Plane): add point on plane query

BotellaA:


src/geode/geometry/basic_objects/plane.cpp
```diff
@@ -74,6 +74,28 @@ namespace geode
...
+        for( const auto d : LRange{ 3 } )
+        {
+            if( std::fabs( normal_.value( d ) ) < GLOBAL_EPSILON )
+            {
+                continue;
+            }
+            const auto d1 = ( d + 1 ) % 3;
+            const auto d2 = ( d + 2 ) % 3;
+            point.set_value( d1, 1 );
+            point.set_value( d2, 1 );
```
panquez: should we set absolute location to one or a delta from origin? what happens if th origin is located at (1, 1, 1)?

src/geode/geometry/basic_objects/plane.cpp
```diff
@@ -74,6 +74,34 @@ namespace geode
...
+        Point3D point;
+        for( const auto direction : LRange{ 3 } )
+        {
+            if( std::fabs( normal_.value( direction ) ) < GLOBAL_EPSILON )
+            {
+                continue;
+            }
+            const auto direction1 = ( direction + 1 ) % 3;
+            const auto direction2 = ( direction + 2 ) % 3;
+            const Point3D& origin = origin_;
```
panquez, line 89: why do you need a local ref of this member?

panquez (approved):

src/geode/geometry/basic_objects/plane.cpp
BotellaA: yes because it can be a point3D or a ref wrapper of a point3D

PR #1294, opened 2026-06-24
fix(Model): Added api functions to use and define internal relations …

MelchiorSchuh:
…between corners and closed lines


BotellaA (approved):
do we need to upgrade the tests ?

PR #1298, opened 2026-07-01
fix(IntersectionDetection): missing segment/segment 3D

BotellaA:


include/geode/geometry/intersection_detection.hpp
panquez: to update

panquez, line 77: also in 3D?

src/geode/geometry/intersection_detection.cpp
```diff
@@ -150,62 +190,62 @@ namespace
...
-            return std::make_pair( POSITION::vertex1, POSITION::inside );
-        }
-        if( s1_p0_side == SIDE::zero )
-        {
-            return std::make_pair( POSITION::inside, POSITION::vertex0 );
-        }
-        if( s1_p1_side == SIDE::zero )
-        {
-            return std::make_pair( POSITION::inside, POSITION::vertex1 );
         }
```
panquez, line 215: re-use best_project_axis ?

panquez (approved):

src/geode/geometry/intersection_detection.cpp
BotellaA: no it uses a triangle

PR #1301, opened 2026-07-07
Feat/2d medial balls

MoulardM:


src/geode/model/helpers/ray_tracing.cpp
```diff
@@ -307,24 +299,20 @@ namespace geode
         }
 
     private:
-        struct CachedTree
-        {
-            absl::once_flag built;
```
BotellaA, line 312: remove also useless includes

BotellaA (approved):

PR #1317, opened 2026-08-12
feat(ModelBuilders): change parameters in  model builder functions

BenPinet:
BREAKING CHANGE: take components instead of uuid  in model builders  such as set_surface_name

### DIFF
--- a/include/geode/model/mixin/builder/surfaces_builder.hpp
+++ b/include/geode/model/mixin/builder/surfaces_builder.hpp
@@ -33,9 +33,7 @@
 namespace geode
 {
     FORWARD_DECLARATION_DIMENSION_CLASS( Surface );
-    ALIAS_2D_AND_3D( Surface );
     FORWARD_DECLARATION_DIMENSION_CLASS( Surfaces );
-    ALIAS_2D_AND_3D( Surfaces );
     FORWARD_DECLARATION_DIMENSION_CLASS( SurfaceMesh );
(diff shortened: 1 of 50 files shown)


include/geode/model/mixin/builder/surfaces_builder.hpp
```diff
@@ -33,7 +33,9 @@
 namespace geode
 {
     FORWARD_DECLARATION_DIMENSION_CLASS( Surface );
+    ALIAS_2D_AND_3D( Surface );
     FORWARD_DECLARATION_DIMENSION_CLASS( Surfaces );
+    ALIAS_2D_AND_3D( Surfaces );
```
panquez, line 38: Why do you need the aliases?

include/geode/model/representation/builder/detail/copy.hpp
```diff
@@ -51,83 +51,90 @@ namespace geode
     {
         using Mapping = ModelCopyMapping::Mapping;
 
-        template < typename ModelFrom, typename BuilderTo >
-        void copy_corner_components(
-            const ModelFrom& from, BuilderTo& builder_to, Mapping& mapping )
+        template < typename ModelFrom, typename ModelTo >
+        void copy_corner_components( const ModelFrom& from,
+            const ModelTo& to,
```
panquez, line 56: "to" is too short for tidy

panquez (approved):

PR #1321, opened 2026-08-14
fix(Segment): make segment.direction() return normalized direction

BenPinet:


src/geode/geometry/distance.cpp
```diff
@@ -486,8 +486,8 @@ namespace
         /* Algorithm and code found on
          * https://github.com/davideberly/GeometricTools/blob/master/GTE/Mathematics/DistSegmentSegment.h
          */
-        const auto P1mP0 = segment0.direction();
-        const auto Q1mQ0 = segment1.direction();
+        const auto P1mP0 = segment0.direction() * segment0.length();
+        const auto Q1mQ0 = segment1.direction() * segment1.length();
```
panquez: maybe better to create a vector (segment1.p0 -> segment1.p1) instead of doing the normalization then the denormalisation

src/geode/geometry/distance.cpp
```diff
@@ -898,7 +898,7 @@ namespace geode
         segment_line_distance( const Segment< dimension >& segment,
             const InfiniteLine< dimension >& line )
     {
-        const auto segDirection = segment.direction();
+        const auto segDirection = segment.direction() * segment.length();
```
panquez: same here

panquez (approved):

PR #1324, opened 2026-08-25
feat(VertexIdentifier): add a unique_vertex_attribute_id

BenPinet:


src/geode/model/helpers/convert_model_meshes.cpp
```diff
@@ -94,15 +96,19 @@ namespace
...
         }
         else if( mesh_type
                  == geode::PolygonalSurface< Model::dim >::type_name_static() )
         {
-            builder.update_surface_mesh( surface,
-                std::move( geode::convert_surface_mesh_into_polygonal_surface(
-                    mesh ) ) );
+            auto poly_surface =
+                geode::convert_surface_mesh_into_polygonal_surface( mesh );
+            poly_surface->vertex_attribute_manager().delete_attribute(
```
panquez, line 109: i do not understand why there is the need of attribute deletion

BenPinet: The converted surface still has the attribute unique_vertices. 
During the update_surface_mesh, we do a register component that creates the attribute unique vertices on the new surface mesh.
If it still exists on the surface,  the creation is impossible because we try to create two attributes with the same id.
Before my changes we just created a new attribute and the surface had two attributes name unique_vertices, with the more recent one being used in the vertex identifier.

src/geode/model/mixin/core/vertex_identifier.cpp
```diff
@@ -217,19 +221,51 @@ namespace geode
...
+                OpenGeodeModelException::check_exception(
+                    unique_vertices_ids.has_value(), nullptr,
+                    OpenGeodeException::TYPE::data,
+                    "[VertexIdentifier::load_component] Unique vertices "
+                    "attribute not found." );
+                OpenGeodeModelException::check_exception(
+                    unique_vertices_ids.value().size() == 1, nullptr,
+                    OpenGeodeException::TYPE::data,
+                    "[VertexIdentifier::load_component] Unique vertices "
+                    "attribute is not unique." );
```
MelchiorSchuh: Verify if attribute type is index_t and give component id in exception message -> maybe encapsulate this in a function to add readability ?

MelchiorSchuh (approved):

src/geode/model/mixin/core/vertex_identifier.cpp
```diff
@@ -217,19 +221,51 @@ namespace geode
...
-                    UNIQUE_VERTICES_NAME );
-            OpenGeodeModelException::check_exception(
-                unique_vertices_ids.has_value(), nullptr,
-                OpenGeodeException::TYPE::data,
-                "[VertexIdentifier::load_component] Unique vertices "
-                "attribute not found." );
+            if( !mesh.vertex_attribute_manager().attribute_exists(
+                    unique_vertex_id_ ) )
+            {
+                // For old files before v18 of OpenGeode the unique vertices
```
BotellaA: Add a function with enough info to try to remove the comment

BotellaA (approved):

panquez (approved):

PR #1329, opened 2026-09-08
fix(attribute_import): Finished removing import methods using Spans

MelchiorSchuh:


BotellaA (approved):
apply tidy comments 🙏

Issue #1194, opened 2025-10-01
Normal computation fallback error with flat triangles.


BenoitTHEBAULT:
The normal computation leads to obtaining an aberrant normal rather than an nullopt.
flat_triangle = [33566.2651233045 25138.4986462893 -2492.99708294131, 33984.3154256702 25601.9798292691 -2496.22014502399, 33775.2894528137 25370.2373676621 -2494.5090680561].

PR #1332, opened 2026-09-10
Fix issue with angularly flat triangles returning anomalous normal

anj1:
This PR fixes an issue where the auxiliary-triangle fallback was causing angularly flat triangles to return an anomalous normal rather than nullopt; see https://github.com/Geode-solutions/OpenGeode/issues/1194


panquez (approved):

BotellaA:
Thanks @anj1 for your contribution! :pray:

PR #1333, opened 2026-09-10
fix(Triangle): better handle flat triangle normal computation

BotellaA:
fix #1194

### DIFF
--- a/include/geode/geometry/basic_objects/triangle.hpp
+++ b/include/geode/geometry/basic_objects/triangle.hpp
@@ -51,7 +51,6 @@ namespace geode
         {
             geode::local_index_t pivot{ geode::NO_LID };
             geode::Vector3D normal{ { 0, 0, 0 } };
-            std::array< double, 3 > lengths;
         };
     } // namespace internal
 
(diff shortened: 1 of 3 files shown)


panquez (approved):

BotellaA:
@anj1 did you really have an issue with this code? I am unsure about issue #1194, which seems to produce a valid normal.

anj1:
I'm interpeting `GLOBAL_ANGULAR_EPSILON` as a threshold for triangle validity, while the fallback appears to treat it only as a direct-pivot conditioning threshold.

So you bring up a good point. Which one of these is correct? If the fallback behavior is intended, maybe it's better to separate`normal()` from strict angular degen detection.

BotellaA:
GLOBAL_ANGULAR_EPSILON is meant to detect too small/large angle between vectors. Those angles make any geometric computation unreliable. So first, try to compute a pivot, a triangle vertex where its angle is good enough. 

Then, if no vertex is found but the triangle is not degenerate (meaning all vertices are not aligned on a thick line), the triangle can be split to generate a smaller triangle where the angles are enough to compute a normal.

BotellaA:
Not sure to follow what you mean by separate the two

anj1:
I think based on your explanation the original code probably implemented the correct semantics then.

BotellaA:
If you have a counter example, do not hesitate to reach out.

